PR #3, opened 2018-01-25
feat: trash module, trashcan handling

didoda:
This provides basic trash module.
Each module can access his own trashcan at `/trash?filter[type]=<object_type>`.
Trashcan provides single objects `restore` and `delete`.

### DIFF
--- a/src/Model/API/BEditaClient.php
+++ b/src/Model/API/BEditaClient.php
@@ -259,7 +259,7 @@ class BEditaClient
      * @param string $type Object type name
      * @return array Response in array format
      */
-    public function restoreObject($id, $type = 'objects')
+    public function restoreObject($id, $type)
     {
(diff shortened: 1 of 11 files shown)


src/Model/API/BEditaClient.php
```diff
@@ -241,6 +241,37 @@ public function deleteObject($id, $type)
...
+    }
+
+    /**
+     * Restore object from trash
+     *
+     * @param int|string $id Object id
+     * @param string $type Object type name
+     * @return array Response in array format
+     */
+    public function restoreObject($id, $type = 'objects')
```
stefanorosanelli: since `objects` is abstract I would not allow empty `$type` argument
BTW: does restore `/objects/{id}` work? :thinking:

PR #4, opened 2018-01-26
feat: flash messages

didoda:
This provides basic flash messages on controllers.
Usage:
```
$this->Flash->info('This is an info');
$this->Flash->warning('This is a warning');
$this->Flash->error('This is an error');
$this->Flash->success('This is a success');
```

### DIFF
--- a/src/Controller/ModulesController.php
+++ b/src/Controller/ModulesController.php
@@ -125,7 +125,7 @@ class ModulesController extends AppController
         $this->apiResponse = $this->apiClient->saveObject($this->objectType, $this->request->getData());
         // TODO: error if $this->apiResponse['data']['id'] empty or  $this->apiResponse['error'] not empty
 
-        $this->Flash->success('Object saved');
+        $this->Flash->success(__('Object saved'));
 
         return $this->redirect(Router::url(sprintf('/%s/view/%s', $this->objectType, $this->apiResponse['data']['id'])));
     }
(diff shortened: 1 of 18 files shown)


stefanorosanelli:
Seems ok!
I would drop `success` for now and use only `info`, `warn`, `error` like in BE3

src/Controller/ModulesController.php
```diff
@@ -125,6 +125,8 @@ public function save()
         $this->apiResponse = $this->apiClient->saveObject($this->objectType, $this->request->getData());
         // TODO: error if $this->apiResponse['data']['id'] empty or  $this->apiResponse['error'] not empty
 
+        $this->Flash->success('Object saved');
```
stefanorosanelli: what about i18n?
```
$this->Flash->success(__('Object saved'));
```
same applies to other messages

src/Controller/TrashController.php
```diff
@@ -83,7 +83,9 @@ public function view($id)
      */
     public function restore()
     {
-        $this->apiResponse = $this->apiClient->restoreObject($this->request->getData('id'));
+        $this->apiResponse = $this->apiClient->restoreObject($this->request->getData('id'), $this->request->getData('object_type'));
+
+        $this->Flash->success('Object restored');
```
stefanorosanelli: I would use info for now, like
```
 $this->Flash->info(__('Object restored'));
```

didoda:
After last commit, usage:
```
$this->Flash->info('This is an info');
$this->Flash->warning('This is a warning');
$this->Flash->error('This is an error');
```

PR #5, opened 2018-01-29
JSON Schema usage

stefanorosanelli:
* configuration moved to `dotenv` as default,
* new `SchemaComponent` to store and handle schema data via simple cache (to be improved)
* `PropertyHelper` to display dynamic properties

Dynamic property tags are in currently in `Element/Property/{name}.twig` but there are many other (better) solutions

### DIFF
--- a/src/Controller/Component/SchemaComponent.php
+++ b/src/Controller/Component/SchemaComponent.php
@@ -1,0 +1,93 @@
+<?php
+/**
+ * BEdita, API-first content management framework
+ * Copyright 2018 ChannelWeb Srl, Chialab Srl
+ *
+ * This file is part of BEdita: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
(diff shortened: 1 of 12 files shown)


src/Controller/Component/SchemaComponent.php
```diff
@@ -0,0 +1,93 @@
...
+        'apiClient' => null,
+    ];
+
+    /**
+     * {@inheritDoc}
+     */
+    public function initialize(array $config)
+    {
+        $this->apiClient = Hash::get($config, 'apiClient', null);
+        $this->type = Hash::get($config, 'type', null);
```
fquffio: `null` is not necessary as third parameter here: it is assumed by default.

In general, everything in this method is not necessary, since these values could be obtained with `$this->getConfig('apiClient')` and `$this->getConfig('type')`, so there's no need to persist these values as instance attributes. Also, this might cause unexpected behaviors if the user invokes `$this->Schema->setConfig()` in the controller, because the values are not sync'ed.

PR #8, opened 2018-02-01
Add many return & static type hints

fquffio:
This is mainly a proposal: adding scalar type hints to method parameters, and return type hints where possible.

### DIFF
--- a/src/Model/API/BEditaClient.php
+++ b/src/Model/API/BEditaClient.php
@@ -191,7 +191,7 @@ class BEditaClient
      * @param array|null $headers Headers
      * @return array|null Response in array format
      */
-    public function get(string $path, array $query = null, array $headers = null) : ?array
+    public function get(string $path, ?array $query = null, ?array $headers = null) : ?array
     {
         $this->sendRequestRetry('GET', $path, $query, $headers);
 
@@ -206,7 +206,7 @@ class BEditaClient
      * @param array|null $headers Custom request headers
      * @return array|null Response in array format
      */
-    public function getObjects(string $type = 'objects', array $query = null, array $headers = null) : ?array
+    public function getObjects(string $type = 'objects', ?array $query = null, ?array $headers = null) : ?array
     {
         return $this->get(sprintf('/%s', $type), $query, $headers);
     }
@@ -220,7 +220,7 @@ class BEditaClient
      * @param array|null $headers Custom request headers
      * @return array|null Response in array format
      */
-    public function getObject($id, string $type = 'objects', array $query = null, array $headers = null) : ?array
+    public function getObject($id, string $type = 'objects', ?array $query = null, ?array $headers = null) : ?array
     {
         return $this->get(sprintf('/%s/%s', $type, $id), $query, $headers);
     }
@@ -233,7 +233,7 @@ class BEditaClient
      * @param array|null $headers Custom request headers
      * @return array|null Response in array format
      */
-    public function saveObject(string $type, array $data, array $headers = null) : ?array
+    public function saveObject(string $type, array $data, ?array $headers = null) : ?array
     {
         $id = null;
         if (array_key_exists('id', $data)) {
@@ -320,7 +320,7 @@ class BEditaClient
      * @param array|null $headers Custom request headers
      * @return array|null Response in array format
      */
-    public function patch(string $path, $body, array $headers = null) : ?array
+    public function patch(string $path, $body, ?array $headers = null) : ?array
     {
         $this->sendRequestRetry('PATCH', $path, null, $headers, $body);
 
@@ -335,7 +335,7 @@ class BEditaClient
      * @param array|null $headers Custom request headers
      * @return array|null Response in array format
      */
-    public function post(string $path, $body, array $headers = null) : ?array
+    public function post(string $path, $body, ?array $headers = null) : ?array
     {
         $this->sendRequestRetry('POST', $path, null, $headers, $body);
 
@@ -365,7 +365,7 @@ class BEditaClient
      * @param string|resource|\Psr\Http\Message\StreamInterface|null $body Request body.
      * @return \Psr\Http\Message\ResponseInterface
      */
-    protected function sendRequestRetry(string $method, string $path, array $query = null, array $headers = null, $body = null) : ResponseInterface
+    protected function sendRequestRetry(string $method, string $path, ?array $query = null, ?array $headers = null, $body = null) : ResponseInterface
     {
         try {
             return $this->sendRequest($method, $path, $query, $headers, $body);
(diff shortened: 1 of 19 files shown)


src/Model/API/BEditaClient.php
```diff
@@ -187,11 +187,11 @@ public function authenticate($username, $password)
...
      * @param string $path Endpoint URL path to invoke
-     * @param array $query Optional query string
-     * @param array $headers Headers
-     * @return array Response in array format
+     * @param array|null $query Optional query string
+     * @param array|null $headers Headers
+     * @return array|null Response in array format
      */
-    public function get($path, $query = [], $headers = [])
+    public function get(string $path, array $query = null, array $headers = null) : ?array
```
didoda: Shouldn't it be the following?
```
public function get(string $path, ?array $query = null, ?array $headers = null) : ?array
```

src/Model/API/BEditaClient.php
```diff
@@ -202,11 +202,11 @@ public function get($path, $query = [], $headers = [])
...
      * @param string $type Object type name
-     * @param array $query Optional query string
-     * @param array $headers Custom request headers
-     * @return array Response in array format
+     * @param array|null $query Optional query string
+     * @param array|null $headers Custom request headers
+     * @return array|null Response in array format
      */
-    public function getObjects($type = 'objects', $query = [], $headers = [])
+    public function getObjects(string $type = 'objects', array $query = null, array $headers = null) : ?array
```
didoda:
```
public function getObjects(string $type = 'objects', ?array $query = null, ?array $headers = null) : ?array
```
If I'm not misunderstanding `?` operator...

src/Model/API/BEditaClient.php
```diff
@@ -216,11 +216,11 @@ public function getObjects($type = 'objects', $query = [], $headers = [])
...
      * @param string $type Object type name
-     * @param array $query Optional query string
-     * @param array $headers Custom request headers
-     * @return array Response in array format
+     * @param array|null $query Optional query string
+     * @param array|null $headers Custom request headers
+     * @return array|null Response in array format
      */
-    public function getObject($id, $type = 'objects', $query = [], $headers = [])
+    public function getObject($id, string $type = 'objects', array $query = null, array $headers = null) : ?array
```
didoda: as above comment:
```
public function getObject($id, string $type = 'objects', ?array $query = null, ?array $headers = null) : ?array
```

src/Model/API/BEditaClient.php
```diff
@@ -230,10 +230,10 @@ public function getObject($id, $type = 'objects', $query = [], $headers = [])
      *
      * @param string $type Object type name
      * @param array $data Object data to save
-     * @param array $headers Custom request headers
-     * @return array Response in array format
+     * @param array|null $headers Custom request headers
+     * @return array|null Response in array format
      */
-    public function saveObject($type, $data, $headers = [])
+    public function saveObject(string $type, array $data, array $headers = null) : ?array
```
didoda: Again:
```
public function saveObject(string $type, array $data, ?array $headers = null) : ?array
```

src/Model/API/BEditaClient.php
```diff
@@ -317,10 +317,10 @@ public function restoreObject($id, $type)
      *
      * @param string $path Endpoint URL path to invoke
      * @param mixed $body Request body
-     * @param array $headers Custom request headers
-     * @return array Response in array format
+     * @param array|null $headers Custom request headers
+     * @return array|null Response in array format
      */
-    public function patch($path, $body, $headers = [])
+    public function patch(string $path, $body, array $headers = null) : ?array
```
didoda:
```
public function patch(string $path, $body, ?array $headers = null) : ?array
```
Ignore comment, if `?array $headers = null` is not the right syntax

src/Model/API/BEditaClient.php
```diff
@@ -332,10 +332,10 @@ public function patch($path, $body, $headers = [])
      *
      * @param string $path Endpoint URL path to invoke
      * @param mixed $body Request body
-     * @param array $headers Custom request headers
-     * @return array Response in array format
+     * @param array|null $headers Custom request headers
+     * @return array|null Response in array format
      */
-    public function post($path, $body, $headers = [])
+    public function post(string $path, $body, array $headers = null) : ?array
```
didoda: As previous comment `?array $headers = null`

src/Model/API/BEditaClient.php
```diff
@@ -365,7 +365,7 @@ public function delete($path)
      * @param string|resource|\Psr\Http\Message\StreamInterface|null $body Request body.
      * @return \Psr\Http\Message\ResponseInterface
      */
-    protected function sendRequestRetry($method, $path, array $query = null, array $headers = null, $body = null)
+    protected function sendRequestRetry(string $method, string $path, array $query = null, array $headers = null, $body = null) : ResponseInterface
```
didoda: `?array $query = null` etc.?

src/Model/API/BEditaClient.php
```diff
@@ -396,7 +396,7 @@ protected function sendRequestRetry($method, $path, array $query = null, array $
      * @return \Psr\Http\Message\ResponseInterface
      * @throws \App\Model\API\BEditaClientException Throws an exception if server response code is not 20x.
      */
-    protected function sendRequest($method, $path, array $query = null, array $headers = null, $body = null)
+    protected function sendRequest(string $method, string $path, array $query = null, array $headers = null, $body = null) : ResponseInterface
```
didoda: ?array $query = null etc.?

src/Model/API/BEditaClient.php
fquffio: This applies to both this and the following comments.

You're right, yet I think this works anyway. Consider the following example:

```php
<?php

function test1(array $foo = null)
{
    var_dump($foo);
}

function test2(array $foo)
{
    var_dump($foo);
}

function test3(?array $foo = null)
{
    var_dump($foo);
}

function test4(?array $foo)
{
    var_dump($foo);
}

foreach (['test1', 'test2', 'test3', 'test4'] as $fun) {
    echo "$fun(null)" . PHP_EOL;
    try {
        $fun(null);
    } catch (\Throwable $e) {
        echo sprintf('ERROR [%s]: %s', get_class($e), $e->getMessage()) . PHP_EOL;
    }
}

/* OUTPUT:

test1(null)
NULL
test2(null)
ERROR [TypeError]: Argument 1 passed to test2() must be of the type array, null given, called in <thisfile> on line 26
test3(null)
NULL
test4(null)
NULL

*/
```

The only error is triggered by `test2(null)`, because its first argument is of type `array` and its default value isn't `null`. `test1(null)` works because its first argument is of type is `array`, but the supplied default value is `null`, thus `null` is an accepted value. `test3` and `test4` accept an argument of type `?array`, thus `null` is acceptable whether it's the default value or not.

So, as far as I can tell `?array $myArr = null` and `array $myArr = null` are completely interchangeable. I find the first to be more correct, and the latter to be less verbose. I'm changing it as you've rightfully suggested, but we should keep this convention in mind in the future, since both will work.

didoda: Well, if both forms are accepted and equivalent, we shouldn't be worried about shorter forms in code, when we eventually forget to use the "explicit/redundant" version `?array $myArr = null` ^_^

Issue #15, opened 2018-02-12
List filters


stefanorosanelli:
Implement first filter: text search - other filters will come later

- https://docs.bedita.net/en/latest/endpoints/parameters.html#search-query-filter - use `filter[query]` or `q` query string
- buttons: 'find` (or `apply`) and `reset filter`
- filter must persist with pagination

PR #19, opened 2018-02-13
Feat/15 list filters

didoda:
This provides basic text filter (#15 )


src/Controller/ModulesController.php
```diff
@@ -79,6 +80,10 @@ public function index() : ?Response
         $this->set(compact('meta'));
         $this->set(compact('links'));
 
+        if (!empty($query['q'])) {
+            $this->set('searchtext', $query['q']);
```
stefanorosanelli: we could avoid passing `searchtext` to view and read from `$this->request->query` directly with a new helper method `Link::query('q')`

this will make it work for `trash` module too, where it's currently missing

PR #20, opened 2018-02-14
feat: show message on login fail

didoda:
This provides a feedback to user on login fail


src/Controller/LoginController.php
```diff
@@ -35,18 +37,27 @@ public function login() : ?Response
...
 
         // Attempted login.
-        $user = $this->Auth->identify();
-        if ($user) {
+        $user = null;
+        $reason = 'Username or password is incorrect';
+        try {
+            $user = $this->Auth->identify();
+        } catch (BEditaClientException $e) {
+            $this->log($e, LogLevel::ERROR);
```
stefanorosanelli: not an `ERROR`, probably not even a `WARNING`, I'll use `INFO` log level for now

PR #24, opened 2018-02-19
feat: json pretty textarea for object content type

didoda:
This provides `pretty` json textarea for object fields with schema type `object`


stefanorosanelli:
Upon every API request, like in `$this->apiClient->saveObject()` we do a `json_encode($body)` where `$body` is the request data body: `extra` must not be a json string at that point but a PHP array instead.

PR #31, opened 2018-02-23
Feat/add relation

didoda:
This provides add relation using `getRelated`, `addRelated` and `removeRelated` `BEdita API php-sdk` methods.

### DIFF
--- a/src/Controller/ModulesController.php
+++ b/src/Controller/ModulesController.php
@@ -65,12 +65,7 @@ class ModulesController extends AppController
         try {
             $response = $this->apiClient->getObjects($this->objectType, $this->request->getQueryParams());
         } catch (BEditaClientException $e) {
-            // no error flash for autocomplete
-            if (!empty($this->request->getQueryParams()['autocomplete'])) {
-                $this->render('autocomplete');
-
-                return null;
-            }
+            // @TODO: ajax error display
 
             // Error! Back to dashboard.
             $this->log($e, LogLevel::ERROR);
--- a/src/Template/Modules/related.twig
+++ b/src/Template/Modules/related.twig
@@ -32,7 +32,7 @@
 
         {# Count of items per module #}
         <div class="results">
-            {{ paging.count }}
+            {{ __('items') }}: {{ paging.count }}
         </div>
 
         {# Pagination data #}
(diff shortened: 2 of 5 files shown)


src/Controller/AppController.php
```diff
@@ -118,5 +118,19 @@ protected function prepareRequest() : void
             }
             unset($this->request->data['_jsonKeys']);
         }
+        // 'cart' for relations view/save - prepare api calls
+        if (!empty($data['cart'])) {
```
stefanorosanelli: `Superminor` note on naming, may be ignored. I'd prefer for clarity:
* `$data['relations`] or `'rel'` or `'related'` instead of `$data['cart']`
* same key used on `$this->request->data` instead of `'api'`

src/Controller/ModulesController.php
```diff
@@ -65,6 +65,13 @@ public function index() : ?Response
         try {
             $response = $this->apiClient->getObjects($this->objectType, $this->request->getQueryParams());
         } catch (BEditaClientException $e) {
+            // no error flash for autocomplete
+            if (!empty($this->request->getQueryParams()['autocomplete'])) {
```
stefanorosanelli: `[minor]` what errors are we catching form `autocomplete` should we hide them?

src/Template/Modules/related.twig
```diff
@@ -1,42 +1,75 @@
...
+            'data-relation-name': relation,
+            'value': Link.query('q')|default(''),
+        })|raw }}
+    </div>
+
+    {# Toolbar pagination #}
+    <div class="toolbar_pagination">
+
+        {# Count of items per module #}
+        <div class="results">
```
stefanorosanelli: If no items are related a simple `No items found` message should be displayed, as in modules empty index case.

Pagination: if the number of items is <= of the minimum available page size:
 * no page size select displayed
 * no pagination `First / Next / Last` links

src/Template/Modules/related.twig
```diff
@@ -1,42 +1,75 @@
...
+            {{ paging.count }} {{ currentModule.name }}
```
stefanorosanelli: I'd go with `{{paging.count}} items` since generally related object types are different from the current one.

src/Controller/ModulesController.php
didoda: No idea of which errors...possible API php-sdk or server errors. Custom error handling provided on our cakephp web app would render an error on dom element target of ajax call, and we don't want that.

stefanorosanelli: ok, we should be able to display errors on ajax call and not hide them => to do in a new issue
I'd probably remove this handling and add a simple comment like: `// ajax calls errors should be displayed....`

PR #33, opened 2018-03-02
Refactor/class naming

vallo87:
- Simplified class naming convention (see wiki, to be translated)
- Template Folder structure changed to
```
-> Template
     -> Element (elements template)
     -> Layout (layouts)
     -> Pages (pages template matching view controllers)
```


src/Controller/AppController.php
```diff
@@ -100,6 +100,8 @@ public function beforeRender(Event $event) : void
 
             $this->set(compact('user'));
         }
+
+        $this->viewBuilder()->setTemplatePath('Pages/' . $this->name);
```
fquffio: This can probably be achieved by overriding Controller's `_viewPath` method, thereby making an explicit call to `setTemplatePath` not needed:

```php
class AppController extends Controller
{
    // ...
    protected function _viewPath()
    {
        return sprintf('Pages/%s', parent::_viewPath());
    }
}
```

src/Template/Pages/Dashboard/index.twig
```diff
@@ -2,9 +2,11 @@
 {% do _view.assign('bodyViewClass', 'view-dashboard') %}
 
 <div class="dashboard">
-    <ul class="dashboard__modules">
+    <ul class="dashboard-items">
     {% for module in modules %}
-        <li class="dashboard__module dashboard__module--{{ module.name }}">
+
+        {# We should comment nearby any variable dependent class either a list of outcomes, when shor, or the logic behind it #}
```
fquffio: > when shor

Typo?

src/Controller/ErrorController.php
```diff
@@ -57,7 +57,7 @@ public function beforeRender(Event $event) : void
         parent::beforeRender($event);
 
         $this->viewBuilder()->setClassName('App\View\AppView');
-        $this->viewBuilder()->setTemplatePath('Error');
+        $this->viewBuilder()->setTemplatePath('Pages/Error');
```
fquffio: (see comment above)

Since `ErrorController extends AppController`, if the advice in `AppController` is followed this becomes unnecessary.

Issue #34, opened 2018-03-02
Users form field order


stefanorosanelli:
Form fields order in user view (existing o new) is misleading.

*Main attributes* here are:
* `username` and `passowrd` - `confirm password` fields
* `name`
* `surname`
* `email`

All other attributes may go in *other attributes*

didoda:
merged

PR #38, opened 2018-03-05
#34 feat: users form fields order

didoda:
This provides ordered fields for `core_properties` form, by module (specifying properly in config).
For instance, in `app.php`:
```
'Modules' => [
  // [...]
  'users' => [
            'core_attributes' => [
                'username',
                'password',
                'confirm-password',
                'name',
                'surname',
                'email',
            ],
        ],
]

### DIFF
--- a/src/Template/Element/Form/core_properties.twig
+++ b/src/Template/Element/Form/core_properties.twig
@@ -1,18 +1,45 @@
 {% set attributes = data.attributes %}
+{% set module = currentModule.name|default('') %}
+{% set core_properties = config('Modules.' ~ module ~ '.core_attributes')|default(['title', 'description', 'uname', 'status']) %}
 
 <div class="tab"><h2>{{ __('Main Properties') }}</h2></div>
 
 <fieldset id="core_properties">
     {{ Form.hidden('id', {'value': object.id})|raw }}
-    {{ Form.control('title')|raw }}
-    {{ Form.control('description', {'type': 'textarea'})|raw }}
-    {{ Form.control('uname')|raw }}
-    {{ Form.control('status', {
-        'type': 'radio',
-        'options': [
-            {'value': 'on', 'text': __('On')},
-            {'value': 'draft', 'text': __('Draft')},
-            {'value': 'off', 'text': __('Off')},
-        ]
-    })|raw }}
+
+
+{% for key, property in core_properties %}
+    {% if property == 'password' %}
+    <div class="password-component">
+        {{ Form.control('password', {
+            'class': 'password',
+            'placeholder': 'new password',
+            'autocomplete': 'new-password',
+            'default': ''
+        })|raw }}
+        {{ Form.control('password', {
+            'label': 'Retype password',
+            'id': 'confirm_password',
+            'name': 'confirm-password',
+            'class': 'confirm-password',
+            'placeholder': 'confirm password',
+            'autocomplete': 'new-password',
+            'default': ''
+        })|raw }}
+    </div>
+    {% elseif property == 'description' %}
+        {{ Form.control('description', {'type': 'textarea'})|raw }}
+    {% elseif property == 'status' %}
+        {{ Form.control('status', {
+            'type': 'radio',
+            'options': [
+                {'value': 'on', 'text': __('On')},
+                {'value': 'draft', 'text': __('Draft')},
+                {'value': 'off', 'text': __('Off')},
+            ]
+        })|raw }}
+    {% else %}
+        {{ Form.control(property, {'type': Schema.getControlTypeFromSchema(property)})|raw }}
+    {% endif %}
+{% endfor %}
 </fieldset>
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: src/View/Helper/ArrayHelper.php [hunk 1]
```diff
@@ -25,10 +25,31 @@ class ArrayHelper extends Helper
      * Return array_combine of array using values as keys.
      *
      * @param array $arr The array.
+     *
```
fquffio: [minor] We usually don't add a blank line between `@param` and `@return`.

src/Template/Element/Form/core_properties.twig
```diff
@@ -1,18 +1,45 @@
...
+{% for key, property in core_properties %}
+    {% if property == 'password' %}
+    <div class="password-component">
+        {{ Form.control('password', {
+            'class': 'password',
+            'placeholder': 'new password',
+            'autocomplete': 'new-password',
+            'default': ''
+        })|raw }}
+        {{ Form.control('password', {
```
fquffio: It is not a good idea to have two `Form.control('password', {})`. This should be something else, like `Form.control('password-confirm', {})`, and `name` should not be passed as an option to this method.

src/Template/Element/Form/core_properties.twig
```diff
@@ -1,18 +1,45 @@
...
+        {{ Form.control('status', {
+            'type': 'radio',
+            'options': [
+                {'value': 'on', 'text': __('On')},
+                {'value': 'draft', 'text': __('Draft')},
+                {'value': 'off', 'text': __('Off')},
+            ]
+        })|raw }}
+    {% else %}
+        {{ Form.control(property, {'type': Schema.getControlTypeFromSchema(property)})|raw }}
```
fquffio: This `Schema.getControlTypeFromSchema()` was supposed to do exactly the job that's being done here in this template using a series of `if / else`. Could this logic be moved there instead?

src/View/Helper/ArrayHelper.php
```diff
@@ -25,10 +25,31 @@ class ArrayHelper extends Helper
...
     {
         return array_combine(array_values($arr), array_values($arr));
     }
+
+    /**
+     * Return array without specified keys.
+     *
+     * @param array $arr  The array.
+     * @param array $keys The keys to remove.
+     *
```
fquffio: [minor] Same here about the blank line.

src/View/Helper/ArrayHelper.php
```diff
@@ -25,10 +25,31 @@ class ArrayHelper extends Helper
...
      */
     public function combine(array $arr) : array
     {
         return array_combine(array_values($arr), array_values($arr));
     }
+
+    /**
+     * Return array without specified keys.
+     *
+     * @param array $arr  The array.
```
fquffio: I think PHPCS complains if there's more than one whitespace between parameter's name and its comment. Are you using the CakePHP's rules set?

src/View/Helper/ArrayHelper.php
didoda: something wrong with PHPCS I'm using ^_^

src/Template/Element/Form/core_properties.twig
didoda: I see. It was done like this in another issue, I just moved the snippet from a twig to another one. BTW I'll amend this part as you suggest ;)

src/View/Helper/ArrayHelper.php
didoda: Using `PSR2`, but I'm going to fix it

src/Template/Element/Form/core_properties.twig
fquffio: Yeah, I see you just moved it into another place… but still I don't like that. 😄

src/Template/Element/Form/core_properties.twig
didoda: mhm, not sure. `getControlTypeFromSchema` return a string type, it doesn't handle "view" logic.
However I'll give it a deeper look, trying to use it, instead of ugly `if / else`

fquffio: Actually this doesn't solve all the cases. Specifically, the "password + password confirm" is a bit of an edge case. But for status and description should work.

The "password + confirm password" can be probably combined into a single control, but this requires additional configuration for CakePHP's `FormHelper`, and it's probably too much beyond the scope of this PR.

fquffio: (Return type of the helper method can be widened to "string|array", so that in some cases, like "status", an associative map is returned.)

src/View/Helper/ArrayHelper.php
```diff
@@ -31,4 +31,23 @@ public function combine(array $arr) : array
...
+    /**
+     * Return array without specified keys.
+     *
+     * @param array $arr The array.
+     * @param array $keys The keys to remove.
+     * @return array The array without keys.
+     */
+    public function removeKeys(array $arr, array $keys) : array
+    {
+        $result = [];
```
stefanorosanelli: `return array_diff_key($arr, array_flip($keys))` should have same effect

Issue #40, opened 2018-03-08
File upload and image creation


stefanorosanelli:
see bedita/php-sdk#3

PR #41, opened 2018-03-13
#40 feat: file upload and media creation

didoda:
This provides #40 features: file upload and image creation.
Media types available for upload (viewed as combo in the upload form) are configurable in `config/app.php`:
```
    /**
     * Upload default settings
     *
     * - models => available models (value => label) to associate to upload
     */
    'Upload' => [
        'models' => [
            'images' => 'Image',
            'audios' => 'Audio',
            'videos' => 'Video',
            'files' => 'File',
        ],
    ],
```


config/app.default.php
```diff
@@ -473,4 +473,18 @@
     'Pagination' => [
         'sizeAvailable' => [10, 20, 50, 100]
     ],
+
+    /**
+     * Upload default settings
+     *
+     * - models => available models (value => label) to associate to upload
+     */
+    'Upload' => [
```
stefanorosanelli: These types can be obtained using API to avoid configuration.
Proposal: let's use `images`, `audio`, `videos` and `files` as default set, in code not in config - later via API

Then looking at `/home` response used to build modules list we may see if these types are actually available and if user has permissions to create/modify looking in `meta.resources.{endpoint}.hints.allow`

src/Controller/ModulesController.php
```diff
@@ -285,4 +285,40 @@ public function relationships($id, $relation)
         $this->set(compact('meta'));
         $this->set(compact('links'));
     }
+
+    /**
+     * Upload a file and store it in a media stream
+     *
+     * @return void
+     */
+    public function upload()
```
stefanorosanelli: What about a simple test case for `ModulesController` ??
If it takes too much time we can skip this...

src/Template/Pages/Modules/view.twig
```diff
@@ -1,5 +1,12 @@
 {% do _view.assign('title', currentModule.name|humanize) %}
 
+{% if ((object.id is empty) and (objectType in ['images', 'audio', 'video'])) %}
+<div class="upload-form">
+    <h1>{{ __('Upload {0}', currentModule.singular) }}</h1>
```
stefanorosanelli: we don't have `currentModule.singular` right now, we could get that info from `/model/object_types`... remove it for now

src/Template/Pages/Modules/view.twig
```diff
@@ -1,5 +1,12 @@
 {% do _view.assign('title', currentModule.name|humanize) %}
 
+{% if ((object.id is empty) and (objectType in ['images', 'audio', 'video'])) %}
```
stefanorosanelli: it's `videos`

src/Controller/ModulesController.php
didoda: skip for now... more tests in separate task, I suggest

PR #45, opened 2018-03-16
chore: vue delimiters proposal

stefanorosanelli:
What about ERB style delimiters `<%  '%>`, used also by underscore.js ?

http://ruby-doc.org/stdlib-2.5.0/libdoc/erb/rdoc/ERB.html
http://underscorejs.org/docs/underscore.html#section-165

widely used and no conflicts with Twig `{{ }}` `{% %}` or CakePHP delimiters `<?= =>`

### DIFF
--- a/webroot/js/config/config.js
+++ b/webroot/js/config/config.js
@@ -2,4 +2,5 @@
 
 Vue.config.devtools = true;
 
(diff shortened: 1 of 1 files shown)


vallo87:
Indeed!

Issue #47, opened 2018-03-20
Import page - upload and load, using filters


didoda:
Basic import page using filters.
Source should be a list of available filters (type / options).
On submit file, system should read file, process it and return a result.

PR #49, opened 2018-03-22
#47 feat: import page

didoda:
This provides #47


src/Controller/ImportController.php
```diff
@@ -0,0 +1,88 @@
...
+     * @return \Cake\Http\Response|null the Response.
+     */
+    public function file() : ?Response
+    {
+        try {
+            // prepare import filter
+
+            $filter = $this->request->getData('filter');
+            if (empty($filter)) {
+                throw new BEditaClientException('Import filter not selected', 500);
```
stefanorosanelli: It is not a BEdita API client related exception, I'd use a simple `BadRequestException` - without a `catch` it should simply not happen

src/Controller/ImportController.php
```diff
@@ -0,0 +1,88 @@
...
+            }
+            $importFilter = new $filter($this->apiClient);
```
stefanorosanelli: probably a `FilterRegistry` is a safer and cleaner solution, without configurations, we can do that later anyway...

src/Controller/ImportController.php
```diff
@@ -0,0 +1,88 @@
...
+            if (empty($importFilter)) {
+                throw new BEditaClientException('Import filter class not found', 500);
```
stefanorosanelli: what about a `RuntimeException` here?

src/Core/Filter/ImportFilter.php
```diff
@@ -0,0 +1,38 @@
...
+    /**
+     * Import function.
+     * Elaborate a file content, using options (if available); return a ImportResult result.
+     *
+     * @param string $filename The file name
+     * @param string $filepath The file path
+     * @param array $options The import options
+     * @return App\Core\Result\ImportResult The result
+     */
+    public function import($filename, $filepath, ?array $options = []) : ImportResult
```
stefanorosanelli: `abstract public function` ? this way implementation in extending classes is mandatory

stefanorosanelli:
Seems ok, we only need a test for `ImportController::file()` method, other tests can be `ignored` or can be done later.

src/Core/Filter/ImportFilter.php
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    /**
+     * Import function.
+     * Elaborate a file content, using options (if available); return a ImportResult result.
+     *
+     * @param string $filename The file name
+     * @param string $filepath The file path
+     * @param array $options The import options
+     * @return App\Core\Result\ImportResult The result
```
stefanorosanelli: missing starting `\` => `\App\Core\Result\ImportResult`

tests/TestCase/Controller/ImportControllerTest.php
```diff
@@ -0,0 +1,170 @@
...
+     */
+    protected $filename = 'test.png';
+
+    /**
+     * Setup test
+     *
+     * @param string $filter The filter class full path.
+     * @return void
+     */
+    public function setup(string $filter = null) : void
```
stefanorosanelli: I'd rename to `setupImport` or similar to avoid confusion/conflict with `TestCase::setUp()` method

Issue #50, opened 2018-03-27
Trash - better usability


didoda:
Desiderata:

 - Mantain filter after click on `Restore` and `Delete`
 - `Empty trash` button (when filter selected, remove only current type, i.e. media, when page is `/trash?filter%5Btype%5D=media`, otherwise remove all objects from trash)

stefanorosanelli:
We also need a _confirm_ before those actions (trash, delete or restore)

PR #51, opened 2018-03-28
Feat/50 trash better usability

didoda:
This provides #50 + TrashController tests

### DIFF
--- a/tests/TestCase/Controller/TrashControllerTest.php
+++ b/tests/TestCase/Controller/TrashControllerTest.php
@@ -180,10 +180,9 @@ class TrashControllerTest extends TestCase
     {
         $id = $this->setupControllerAndData(false);
         $this->Trash->restore();
-        $expected = new BEditaClientException('[404] Record not found in table "objects"', 404);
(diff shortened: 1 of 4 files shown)


tests/TestCase/Controller/TrashControllerTest.php
```diff
@@ -0,0 +1,272 @@
...
+    /**
+     * Setup api client and auth
+     *
+     * @return void
+     */
+    private function setupApi() : void
+    {
+        $apiBaseUrl = getenv('BEDITA_API');
+        $apiKey = getenv('BEDITA_API_KEY');
+        $this->client = new BEditaClient($apiBaseUrl, $apiKey);
```
stefanorosanelli: This is already done in `AppController::initialize()` what about a `public getApiClient()` method to get client instead of a `setApiClient()`?
Actually..  `AppController::getApiClient()` could be used in components too....

didoda: In this case we had to force to use another `apiClient` (local) in test context, that's why `setApiClient` could be useful... just for unit test

Issue #43, opened 2018-03-14
Show included `stream` in single item view


stefanorosanelli:
single object view should display included stream, if present 
call to `GET /images/{id}` may include also:

```json
 "included": [
        {
            "id": "b1f3e797-558d-4310-a9dc-f0c19927eb85",
            "type": "streams",
            "attributes": {
                "file_name": "my-image.png",
                "mime_type": "image/png"
            },
            "meta": {
                "version": 1,
                "file_size": 122927,
                "hash_md5": "23de017e93556d45f5b69dd508a868f7",
                "hash_sha1": "74fa37b617a0daad926c68d66fc451a246943ae5",
                "width": null,
                "height": null,
                "duration": null,
                "created": "2018-03-14T08:25:54+00:00",
                "modified": "2018-03-14T08:25:54+00:00",
                "url": "http://example.com/files/b1f3e797-558d-4310-a9dc-f0c19927eb85-doorway-platform.png"
            }
        }
    ]
```
`attributes` and `meta` shoud be displayed `read-only` they're not chengeable
link to image and image itself should be displayed.
[https://github.com/bedita/php-sdk/issues/10 will provide thumb generation, soon]

didoda:
Merged

PR #53, opened 2018-03-30
#43 feat: view image thumb on single image view

didoda:
This implements #43 
In image view, image is shown (thumb created... if some error on creation, use a default gif for missing image).
This uses php-sdk api `thumb` method.


src/Controller/ModulesController.php
```diff
@@ -109,11 +109,12 @@ public function view($id) : ?Response
             return $this->redirect(['_name' => 'modules:list', 'object_type' => $this->objectType]);
         }
 
-        $object = $response['data'];
         $revision = Hash::get($response, 'meta.schema.' . $this->objectType . '.revision', null);
         $schema = $this->Schema->getSchema($this->objectType, $revision);
 
-        $this->set(compact('object', 'schema'));
+        $object = $response['data'];
+        $included = (!empty($response['included'])) ? $response['included'][0] : null;
```
stefanorosanelli: it's probably safer to use `$response['included']` complete array, even though it will usually be just one element

src/Template/Element/Form/included.twig
```diff
@@ -0,0 +1,21 @@
+{% if included %}
+{% set imageattributes = { 0: 'url' } %}
```
stefanorosanelli: cycle through `included` array (see above) checking that `type` is `streams`
or use only first element

tests/TestCase/View/Helper/ThumbHelperTest.php
```diff
@@ -0,0 +1,162 @@
...
+        ];
+    }
+
+    /**
+     * Test `url()` method.
+     *
+     * @dataProvider urlProvider()
+     * @covers ::url()
+     * @param array $input The input array.
+     * @return void
```
stefanorosanelli: missing `@param` for `$expected` - we may use a simple boolean

src/Controller/ModulesController.php
didoda: When we have more than one element, should we show multiple streams/thumbs? it makes sense

Issue #59, opened 2018-04-09
bulk operations in modules index


didoda:
In module index page, it should possible perform bulk operations on elements, i.e.:
 - move to trash
 - change status (on/off/draft)
 - csv export (selected|all) ?

PR #65, opened 2018-04-18
#59 feat: bulk actions

didoda:
This provides basic bulk actions: `Export` (csv), `Change status`, `Delete`.


src/Controller/ModulesController.php
```diff
@@ -334,4 +365,104 @@ public function upload()
             'id' => $response['data']['id'],
         ]);
     }
+
+    /**
+     * Export data in csv format
```
stefanorosanelli: Comment typo

src/Controller/ModulesController.php
```diff
@@ -334,4 +365,104 @@ public function upload()
...
+
+        return $this->redirect(['_name' => 'modules:list', 'object_type' => $this->objectType]);
+    }
+
+    /**
+     * Export data in csv format
+     *
+     * @return void
+     */
+    public function export() : void
```
stefanorosanelli: Why not put this in an `ExportController`? Since we have already an `ImportController` it makes sense.

src/Controller/ModulesController.php
```diff
@@ -334,4 +365,104 @@ public function upload()
...
+    {
+        $this->request->allowMethod(['post']);
+        $fields = ['id', 'title', 'description', 'uname', 'status', 'body', 'lang', 'extra'];
```
stefanorosanelli: I'd fetch all properties this way 
```php
$schema = $this->apiClient->schema($this->objectType);
$fields = array_keys($schema['properties'));
```

src/Controller/ModulesController.php
```diff
@@ -334,4 +365,104 @@ public function upload()
...
+            $response = $this->apiClient->getObjects($this->objectType, ['filter' => ['id' => $ids]]);
+            $csv .= $this->csvRows($response, $fields, $delimiter);
+            $csv = rtrim($csv, "\n"); // remove last newline
+            $this->set('csv', $csv);
+        } else { // export all
+            $response = $this->apiClient->getObjects($this->objectType, [
+                'page' => 1,
+                'page_size' => 100,
+            ]);
+            while (!empty($response['links']['next'])) {
```
stefanorosanelli: I'm not sure of the consequences of an `unlimited` export loop.
We may use just `'ids'` for now and think about a limit later.

src/Template/Pages/Modules/index.twig
```diff
@@ -48,6 +57,28 @@
                     {{ __('No items found') }}
                 {% endfor %}
             </div>
+            {% if objects %}
+            <h2>{{ __('Bulk actions') }}</h2>
+            <div>
+                {{ Form.create(null, {'id': 'form-export', 'url': {'_name': 'modules:export', 'object_type': objectType}})|raw }}
```
stefanorosanelli: enable export only if `currentModule.hints.multiple_types` is `false`

src/Controller/ModulesController.php
```diff
@@ -334,4 +365,104 @@ public function upload()
...
+    {
+        if (empty($response) || empty($response['data'])) {
+            return '';
+        }
+        $csv = '';
+        foreach ($response['data'] as $index => $data) {
+            $cells = [];
+            foreach ($fields as $field) {
+                if (!empty($data[$field])) {
+                    $cells[] = '"' . preg_replace('/"/', '""', $data[$field]) . '"';
```
stefanorosanelli: Why not use `http://php.net/manual/es/function.fputcsv.php` to a temp file?
Ordered fields array may be extracted with ```array_intersect_key($data['attributes'], array_flip($fields))```

src/Controller/ModulesController.php
didoda: For now, only "checked" items export is allowed, via post. I'll drop "unlimited" items export part...

didoda:
Still todo: phpunit tests for ExportController...

stefanorosanelli:
Seems ok.

We better leave `Restore / Remove` buttons in trash index list for each row, also `Delete` button in objects index for each row. Makes sense for UX.

PR #77, opened 2018-04-19
Merge styles

xho:


src/Template/Modules/view.twig
```diff
@@ -0,0 +1,31 @@
+{% do _view.assign('title', currentModule.name|humanize) %}
```
stefanorosanelli: this file is not used anymore... `src/Template/Pages/Modules/view.twig` is used instead

PR #78, opened 2018-04-19
Remove twig cache in `cache clear_all`

stefanorosanelli:
Cache shell has been extended: calling `bin/cake cache clear_all` also twig cache is cleared.

### DIFF
--- a/src/Shell/CacheShell.php
+++ b/src/Shell/CacheShell.php
@@ -30,7 +30,7 @@ class CacheShell extends BaseCacheShell
         $twigCachePath = CACHE . 'twigView';
         $folder = new Folder($twigCachePath);
         if (file_exists($twigCachePath) && !$folder->delete()) {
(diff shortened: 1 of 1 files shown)


src/Shell/CacheShell.php
```diff
@@ -0,0 +1,40 @@
...
+    /**
+     * {@inheritDoc}
+     */
+    public function clearAll()
+    {
+        parent::clearAll();
+        $twigCachePath = CACHE . 'twigView';
+        $folder = new Folder($twigCachePath);
+        if (file_exists($twigCachePath) && !$folder->delete()) {
+            $this->error("Error removing Twig cache files in $twigCachePath");
```
didoda: Scrutinizser suggests as follows:
```
The function Cake\Console\Shell::error() has been deprecated: 3.2.0 Use Shell::abort() instead.
Add a line note
```

didoda: Fixed wih commit

PR #79, opened 2018-04-23
Use env() fallback to setup API client

stefanorosanelli:
This PR adds `env` vars as fallback for `API` configuration: if config `API` keys are not set `env` 
 vars are used.


fquffio:
The previous pattern looked a bit better to me: env vars might not be trivial to set up in some scenarios, plus I don't quite like having `env('SOMETHING')` calls around in the codebase.

stefanorosanelli:
Using `config/.env` env vars setup is trivial and this is the official way to setup the API (see `README`)

I'd probably use a configuration only to override the environment. 
This logic can stay in `ApiClientProvider` class: if a config `API` key is found it is also used, but it may be absent as default.

What do you think?

fquffio:
It's great to have configuration live in the environment by default. However, I'd leave the responsibility to read the environment variables to the `app.php`. This is basically what happens for database configuration (see `env('DATABASE_URL', null)`).

fquffio:
How about something like this in `config/bootstrap.php`?

```php
if (!Configure::check('API')) {
    Configure::write('API', [
        'apiBaseUrl' => env('BEDITA_API'),
        'apiKey' => env('BEDITA_API_KEY'),
    ]);
}
```

It would solve the same problem, and we would keep the name of the environment variables written in a couple of places only, so if we want to rename them later we can do without too much hassle.

PR #80, opened 2018-04-23
Redirect on login

stefanorosanelli:
This PR will allow proper `redirect` on login.

E.g. when user session is terminated user is redirected to an url like `/login?redirect=/documents`.
After a successful login user is now properly redirected to `/documents` as expected.

### DIFF
--- a/src/Controller/LoginController.php
+++ b/src/Controller/LoginController.php
@@ -34,7 +34,6 @@ class LoginController extends AppController
 
         if (!$this->request->is('post')) {
             // Display login form.
-            $this->set('redirect', Hash::get($this->request->getQueryParams(), 'redirect'));
 
             return null;
         }
--- a/src/Controller/LoginController.php
+++ b/src/Controller/LoginController.php
@@ -14,6 +14,7 @@ namespace App\Controller;
 
 use BEdita\SDK\BEditaClientException;
 use Cake\Http\Response;
+use Cake\Utility\Hash;
 use Psr\Log\LogLevel;
 
 /**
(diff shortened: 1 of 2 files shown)


src/Controller/LoginController.php
```diff
@@ -33,6 +34,8 @@ public function login() : ?Response
 
         if (!$this->request->is('post')) {
             // Display login form.
+            $this->set('redirect', Hash::get($this->request->getQueryParams(), 'redirect'));
```
fquffio: `$this->request->getQuery('redirect')` should do the job without using `Hash` explicitly — it's still used, but internally.

fquffio: PS: Maybe it's possible to use `_view.request.getQuery('redirect')` directly in the template.

stefanorosanelli: Right, but on the doc block of `getQuery` I see:
```    
 * Developers are encouraged to use getQueryParams() when possible as it is PSR-7 compliant, and this method
 * is not.
 *
 * ### PSR-7 Alternative
 *
 * $value = Hash::get($request->getQueryParams(), 'Post.id', null);
 *
```

fquffio: Uhm… right. However I can see no deprecation notices or whatever. I'd say that this warning is mostly addressed to users developing a middleware that's meant to be compatible with other PHP frameworks and libraries.

If it's possible to use `_view.request.getQuery('redirect')` directly in the template, I'd go with that. Otherwise, if left in the controller, it's fine like this.

stefanorosanelli: ok, done in twig

PR #83, opened 2018-04-30
fix: save users empty password

didoda:
This fixes users save on empty password (do not change password, in this case)

### DIFF
--- a/src/Controller/AppController.php
+++ b/src/Controller/AppController.php
@@ -121,6 +121,7 @@ class AppController extends Controller
             foreach ($keys as $key) {
                 $data[$key] = json_decode($data[$key]);
             }
+            unset($data['_jsonKeys']);
         }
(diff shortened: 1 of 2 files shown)


src/Controller/AppController.php
```diff
@@ -102,19 +102,25 @@ public function beforeRender(Event $event) : void
...
+        }
+
         if (!empty($data['_jsonKeys'])) {
             $keys = explode(',', $data['_jsonKeys']);
             foreach ($keys as $key) {
-                $this->request->data[$key] = json_decode($data[$key]);
+                $data[$key] = json_decode($data[$key]);
             }
-            unset($this->request->data['_jsonKeys']);
         }
```
stefanorosanelli: add `unset($data['_jsonKeys']);` ?

PR #86, opened 2018-05-02
Improve stream details

stefanorosanelli:
In this PR:

 * display provider thumbnail (usually video) if available in items list
 * embed provider video if available
 * display stream fields like metadata, without input

### DIFF
--- a/src/Template/Element/Modules/thumb.twig
+++ b/src/Template/Element/Modules/thumb.twig
@@ -1,6 +1,12 @@
 {# thumb #}
 
-{% if object.type ==  'images' %}
+{# Use external provider thumbnail if available #}
+{% if object.attributes.provider_thumbnail %}
+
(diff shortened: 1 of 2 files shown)


src/Template/Element/Modules/thumb.twig
```diff
@@ -1,6 +1,12 @@
 {# thumb #}
 
-{% if object.type ==  'images' %}
+{# Use external provider thumbnail if available #}
+{% if object.attributes.provider_thumbnail %}
+
+        <img src={{ object.attributes.provider_thumbnail }} width="50" />
```
didoda: At line 11 there's a `{% set width = 50 %}`, to have a `width` variable. I'd use it here as well... So we could move `{% set width = 50 %}` to line 2... and then change line 6 as follows:
```
<img src={{ object.attributes.provider_thumbnail }} width={{ width }} />
```

PR #90, opened 2018-05-09
Issues/69 parent relation for folders

vallo87:
Folder's view now handles single parent select with tree-view component in single-choice mode1<


src/Controller/ModulesController.php
```diff
@@ -366,6 +366,12 @@ public function relationshipsJson($id, string $relation) : void
             }
 
             $response = $this->apiClient->get($available);
+
+            foreach ($response['data'] as $index => $obj) {
+                if ($obj['id'] == $id) {
+                    array_splice($response['data'], $index, 1);
```
didoda: Why removing data from `$response['data']`? a comment would help in here ;)

webroot/js/app/components/tree-view/tree-view.js
```diff
@@ -131,6 +142,35 @@ Vue.component('tree-view', {
             this.pendingRelations = this.pendingRelations.filter(pending => pending.id !== related.id);
         },
 
+        /**
+         * remove all related objets from prendingRelations Array
```
didoda: typo: `objects'. `prendingRelations` or `pendingRelations`?

webroot/js/app/components/tree-view/tree-view.js
```diff
@@ -115,11 +126,11 @@ Vue.component('tree-view', {
         },
 
         /**
-         * remove an object from prendingrelations Array
+         * remove an object from prendingRelations Array
```
didoda: `prendingRelations` or `pendingRelations`?

src/Controller/ModulesController.php
vallo87: right, i'll comment it (anyway is cause in folder tree i want to remove the element at index $index, which is the current object)

PR #95, opened 2018-05-11
feat: handle enum field nullable in schema helper

didoda:
This fixes a problem with SchemaHelper, handling enum field nullable.


src/View/Helper/SchemaHelper.php
```diff
@@ -171,6 +171,14 @@ public function controlOptions($name, $value, $schema) : array
                 'checked' => $value,
             ];
         } elseif ($type === 'enum') {
+            if (!empty($schema['oneOf'])) {
+                foreach ($schema['oneOf'] as $one) {
+                    if (!empty($one['enum'])) {
+                        $schema['enum'] = ['-1' => ''] + $one['enum'];
```
stefanorosanelli: are we sure about this?

didoda: `oneOf` for an enum... I guess it means that there are only 2 values, the first is `null`, the second is the list of enum values.
I couldn't say if there are other cases.

stefanorosanelli: Sorry, I was referring to `['-1' => '']` to set `null` enum item

stefanorosanelli:
@didoda I did a small change: use `array_unshift` and creted a test
if ok we can merge

Still to solve: on the API side, if value is empty NULL should be forced in these cases

Issue #93, opened 2018-05-11
Introduce EmbedMedia helper


stefanorosanelli:
Used in `media` object types view:

* if object is remote with embedded HTML player -> display the player
* if an image with a stream create a thumbnail with `ThumbHelper`

PR #99, opened 2018-05-12
[WIP] Media helper #93

stefanorosanelli:
This PR introduces `MediaHelper` to view/embed multimedia elements in a page using HTML tags.

* there is only a `view` method, using media object, `included` data and an option array as input like
```
 {{ Media.view(object, included, {'options': {'w': 600}, 'mediaUrl': true})|raw }}
```
where `object` is media data array, `included` is the included part of the response containing `streams`, and `options` array may contain as keys:

  - `providerThumb` (default false) boolean flag, display external provider thumbnail, if available
  - `preset`, string value, thumbnail preset to use
  - `options`, array value, thumbnail options to use including `w` (width) and `h` (height) - ignored if `preset` is set

`ThumbHelper` has been refactored: `ERROR` status now replaces `NOT_AVAILABLE` and `NO_URL` since both cases represent a bad response from the API that should never occurr


stefanorosanelli:
close as obsolete (for now)

PR #105, opened 2018-05-15
Element override from a plugin via configuration

stefanorosanelli:
Using a configuration like this:
```php
    'Elements' => [
        'documents' => [
            'Form/included' => 'MyPlugin',
        ],
    ],
```
We are able to override an element: if we're in `documents` module we can use `MyPlugin.Form/included` instead of `Form/included`

Is there a better way to do this with Cake?

### DIFF
--- a/src/View/AppView.php
+++ b/src/View/AppView.php
@@ -13,6 +13,8 @@
 namespace App\View;
 
 use App\View\Twig\BeditaTwigExtension;
+use Cake\Core\Configure;
+use Cake\Utility\Hash;
 use WyriHaximus\TwigView\View\TwigView;
 
 /**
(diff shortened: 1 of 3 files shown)


src/View/AppView.php
```diff
@@ -49,4 +51,22 @@ public function initialize() : void
...
+     * {@inheritDoc}
+     *
+     * Retrieve custom view `element` from configuration mappings.
+     *
+     * If `Elements.{module_name}.{element_name}` exists in configuration a custom element is loaded
+     */
+    protected function _getElementFileName($name, $pluginCheck = true)
+    {
+        $module = $this->get('currentModule');
+        $custom = Configure::read(sprintf('Elements.%s', Hash::get($module, 'name', '')));
```
batopa: Are we sure `currentModule` is always defined and is an array? `Hash::get()` require the first argument to be an array or  instance of `\ArrayAccess`.

Maybe better `$module = $this->get('currentModule', []);` to have a default.

tests/TestCase/View/AppViewTest.php
```diff
@@ -0,0 +1,63 @@
...
+     */
+    public function testInitialize() : void
+    {
+        $View = new AppView();
+        $extensions = $View->getTwig()->getExtensions();
+        static::assertNotEmpty($extensions);
+    }
+
+    /**
+     * Test `initialize` method
```
batopa: we are testing another method not `initialize`

src/View/AppView.php
stefanorosanelli: It's an array, but it could be `null` - more checks are needed, including the one you're proposing

tests/TestCase/View/AppViewTest.php
stefanorosanelli: It's a silly test just to cover `initialize()`, I just added another assert.
Since `initialize()` is called by constructor I think it's the only way to test it.

PR #113, opened 2018-05-30
Feat/112 panel pagination filter

didoda:
This provides pagination and search filter on relations modal panel

Note: launch `composer update` before testing


src/Controller/ModulesController.php
```diff
@@ -318,11 +317,10 @@ public function relationshipsJson($id, string $relation) : void
                     $available = '/folders';
                     break;
                 default:
-                    $response = $this->apiClient->get($path, $this->request->getQueryParams());
+                    $response = $this->apiClient->get($path, ['page' => 1]); // page 1: we need just the available link from response
```
batopa: Maybe `page_size`?

src/Template/Layout/js/app/components/relation-view/relations-add.js
```diff
@@ -20,42 +20,101 @@ export default {
...
+                this.loadObjects();
+            }
         }
     },
 
     methods: {
+        /**
+         * Return data for panel.
+         *
+         * @return {Object} The data for panel
```
batopa: `@return {void}` ?

Issue #115, opened 2018-06-05
Handle type `array` schema


stefanorosanelli:
A JSON schema like the one provided below must be handled with checkboxes or other control in object view form.

```
{
  "type": "array",
  "uniqueItems": true,
  "items": {
      "type": "string",
      "enum": ["A", "B", "C", "D"]
  }
}
```

We don't need a general purpose `array` handler for now. 
A simple string list case, like the one above, is enough.

PR #116, opened 2018-06-05
feat: handle schema type array

didoda:
This provides basic handling for schema type array (#115), as multiple checkbox list.

### DIFF
--- a/src/View/Helper/SchemaHelper.php
+++ b/src/View/Helper/SchemaHelper.php
@@ -175,7 +175,7 @@ class SchemaHelper extends Helper
                             function ($value) {
                                 return ['value' => $value, 'text' => Inflector::humanize($value)];
                             },
-                            Hash::extract($one, 'items.enum')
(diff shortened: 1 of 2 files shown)


src/View/Helper/SchemaHelper.php
```diff
@@ -165,6 +166,27 @@ public function controlOptions($name, $value, $schema) : array
...
         } elseif ($type === 'checkbox') {
+            if (!empty($schema['oneOf'])) {
+                $options = [];
+                foreach ($schema['oneOf'] as $one) {
+                    if (!empty($one['type']) && ($one['type'] === 'array')) {
+                        $options = array_map(
+                            function ($value) {
+                                return ['value' => $value, 'text' => Inflector::humanize($value)];
+                            },
+                            $one['items']['enum']
```
stefanorosanelli: `$one['items']['enum']` may not exist
I'll add a check, something like
`Hash::extract($one, 'items.enum')`

Issue #73, opened 2018-04-18
Profile page


stefanorosanelli:
Page with logged user data from `User: ###` on left menu

User should be able to change properties like:
 * `password` - with `password confirm`
 * `name`, `surname`, `email`.... and other attributes like `city`, `address` if available

Unchangeable properties:
* `username`
* `uname`
* `status`

Current `roles` can be displayed but not changed

PR #121, opened 2018-06-07
feat: user profile page

didoda:
This provides user profile page (#73).
Bonus: rearrange `default.twig` using `LayoutHelper` methods, to customize viewable parts.

- primarySidebar
- secondarySidebar
- layoutHeader
- layoutContent
- layoutFooter
- messages

When new pages (not module) will be created, `LayoutHelper` would be the place where customize view logic.
Note: using helper `LayoutHelper`, `Import` page gained a better view.


src/Controller/UserProfileController.php
```diff
@@ -0,0 +1,74 @@
...
+     *
+     * @return void
+     */
+    public function view() : void
+    {
+        $this->request->allowMethod(['get']);
+
+        $id = $this->Auth->user('id');
+        try {
+            $response = $this->apiClient->getObject($id, 'users');
```
stefanorosanelli: Here we should use `GET /auth/user`
https://docs.bedita.net/en/latest/endpoints/auth.html#who-am-i

`/users` endpoint may not be accessibile or read-only

src/Controller/UserProfileController.php
```diff
@@ -0,0 +1,74 @@
...
+        if (empty($data['password'])) {
+            unset($data['password']);
+            unset($data['confirm-password']);
+        } elseif ($data['password'] != $data['confirm-password']) {
+            $this->Flash->error(__('Invalid data: password and confirm-password do not match'));
+
+            return $this->redirect(['_name' => 'profile:view']);
+        }
+        try {
+            $response = $this->apiClient->saveObject('users', $data);
```
stefanorosanelli: Same as above: we should use `PATCH /auth/user`
https://docs.bedita.net/en/latest/endpoints/auth.html#update-user-profle

We have to think about password change use case a little bit, not yet supported.

src/View/Helper/SiteHelper.php
```diff
@@ -0,0 +1,124 @@
...
+    /**
+     * Return profile attributes (and value) for specified field(s) group
+     *
+     * @param array $attributes The profile attributes
+     * @param string $group The field(s) group
+     * @return array The filtered profile attributes
+     */
+    public function profileFields($attributes, $group = 'Core') : array
+    {
+        if ($group === 'Change password') {
```
stefanorosanelli: Why not use `PropertiesComponent::viewGroups()` to group fields?

src/View/Helper/SiteHelper.php
```diff
@@ -0,0 +1,124 @@
...
+ */
+namespace App\View\Helper;
+
+use Cake\Utility\Hash;
+use Cake\View\Helper;
+
+/**
+ * Helper for site (utils and config)
+ */
+class SiteHelper extends Helper
```
stefanorosanelli: What about a `LayoutHelper` instead?

src/Template/Pages/UserProfile/view.twig
```diff
@@ -0,0 +1,55 @@
...
+                    {% if key != '0' %}
+                    <header>
+                        <h2>{{ __(group) }}</h2>
+                    </header>
+                    {% endif %}
+                    <div class="tab-container">
+                        {% set jsonKeys = [] %}
+                        {% set attributes = Site.profileFields(object.attributes, group) %}
+                        {% for key, value in attributes %}
+                            {% set options = Schema.controlOptions(key, value, schema.properties[key]) %}
```
stefanorosanelli: Is there a `schema` loaded via controller?

src/Controller/UserProfileController.php
```diff
@@ -0,0 +1,74 @@
...
+        if (empty($data['password'])) {
+            unset($data['password']);
+            unset($data['confirm-password']);
+        } elseif ($data['password'] != $data['confirm-password']) {
+            $this->Flash->error(__('Invalid data: password and confirm-password do not match'));
+
+            return $this->redirect(['_name' => 'profile:view']);
+        }
+        try {
+            $this->apiClient->saveObject('users', $data);
```
stefanorosanelli: Same as above: we should use `PATCH /auth/user`
https://docs.bedita.net/en/latest/endpoints/auth.html#update-user-profle

We have to think about password change use case a little bit, not yet supported.

src/Template/Pages/UserProfile/view.twig
didoda: Nope. I add some logic in controller, like
```
$revision = Hash::get($response, 'meta.schema.users.revision', null);
$schema = $this->Schema->getSchema('users', $revision);
$this->set('schema', $schema);
```

PR #123, opened 2018-06-21
feat: country codes with league/iso3166

didoda:
This provides country iso codes using league/iso3166 (https://github.com/thephpleague/iso3166).
Core model `Countries` and test unit `CountriesTest`.


didoda:
Not necessary.

Issue #124, opened 2018-06-28
Password change in user profile


stefanorosanelli:
After bedita/bedita#1501 password change is allowed if the current valid password is also provided

Instead `username` and `email` should be readonly for now-

PR #130, opened 2018-07-19
Issue/124 password change

vallo87:
#124 
TODO: Validation

suggestion: Vuelidate


src/View/Helper/SchemaHelper.php
```diff
@@ -100,6 +100,28 @@ protected function customControlOptions($name) : array
                     ],
                 ];
 
+            case 'username':
```
stefanorosanelli: `username` not necessary an `email`, I would not represent it as email in general

stefanorosanelli: should not also be `readonly` in general

src/View/Helper/SchemaHelper.php
```diff
@@ -100,6 +100,28 @@ protected function customControlOptions($name) : array
...
+                return [
+                    'class' => 'email',
+                    'readonly' => true,
+                ];
+
+            case 'email':
```
stefanorosanelli: not `readonly` in general `email ` is used in `users` and `profiles` where it may be changed

src/View/Helper/SchemaHelper.php
```diff
@@ -100,6 +100,28 @@ protected function customControlOptions($name) : array
...
+                return [
+                    'class' => 'email',
+                    'readonly' => true,
```
stefanorosanelli: not `readonly` in general `email ` is used in `users` and `profiles` where it may be changed

src/View/Helper/SchemaHelper.php
stefanorosanelli: `username` not necessary an `email`
should not also be `readonly` in general

vallo87: corrected!

src/View/Helper/SchemaHelper.php
vallo87: corrected!

PR #132, opened 2018-07-19
Issue/131 folder tree bug

vallo87:
tree entry indentation was a style issue


src/Template/Element/Form/trees.scss
```diff
@@ -34,7 +34,7 @@
 
     .tree-entry-list {
         padding: 0 $gutter * .5;
-        margin-left: - ($gutter * 1.5);
+        // margin-left: - ($gutter * 1.5);
```
stefanorosanelli: is there a reason to keep this line?

PR #137, opened 2018-07-27
Issue/72 alert on leave page

xho:


src/Template/Layout/js/app/app.js
```diff
@@ -262,6 +270,35 @@ const _vueInstance = new Vue({
...
+         */
+        alertBeforePageUnload() {
+            var forms = [...document.querySelectorAll('form')];
+            forms.forEach((form) => {
+                form.addEventListener('change', () => {
+                    form.changed = true;
+                });
+                form.addEventListener('submit', (ev) => {
+                    if (form.action.endsWith('/delete')) {
+                        if (!confirm("Do you really want to trash the object?")) {
```
stefanorosanelli: How do we handle message localization in js?
In php/twig we use `__()` function for that...

PR #139, opened 2018-08-01
Plugin modules

stefanorosanelli:
Custom modules may be defined in plugins via configuration key `'PluginModules'`.
Usually this configuration is written by plugin bootstrap.

```php
    'PluginModules' => [
        // unique name
        'My Module' => [
            'title' => 'My Module',
            // main module url
            'route' => Router::url(['_name': 'my_module:index']),
            // css class
            'class' => [
                'dashboard' => 'has-background-black icon-sample',
                'menu' => 'has-background-black',
            ],
        ],
    ],
```
These modules are then displayed in dashboard and top-level menu.

### DIFF
--- a/tests/TestCase/View/Helper/LayoutHelperTest.php
+++ b/tests/TestCase/View/Helper/LayoutHelperTest.php
@@ -79,11 +79,6 @@ class LayoutHelperTest extends TestCase
                 null,
                 true,
             ],
-            'currentModule empty' => [
-                null,
-                null,
-                false,
-            ],
             'currentModule not empty' => [
                 null,
                 'Objects',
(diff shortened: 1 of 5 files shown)


tests/TestCase/View/Helper/LayoutHelperTest.php
```diff
@@ -193,11 +192,10 @@ public function layoutFooterProvider() : array
      * @dataProvider layoutFooterProvider()
      * @covers ::layoutFooter()
      */
-    public function testLayoutFooter($currentModule, $expected) : void
+    public function testLayoutFooter($name, $expected) : void
```
batopa: You should change the param annotation to `$name` too.

PR #149, opened 2018-08-03
Fix `json` formatting in all elements

stefanorosanelli:
This PR fixes a bug with `json` input fields.

When a `json` property is used in elements other than `advanced` JSON formatting via `_jsonKeys` is not working as expected.

 * a `_jsonKeys` configuration is used to easily share and update keys
 * a new `write_config` Twig function has been introduced

TODO: there's a formatting issue, probably caused by moving `{{ Form.control('_jsonKeys',...) }}` stil to be fixed.

### DIFF
--- a/src/Template/Pages/Modules/view.twig
+++ b/src/Template/Pages/Modules/view.twig
@@ -59,7 +59,7 @@
             {% element 'Form/relations' %}
 
         {# Set `_jsonKeys` hidden input from config #}
-        {{ Form.control('_jsonKeys', {'type': 'hidden', 'value': config('_jsonKeys')|join(',')})|raw }}
+        {{ Form.control('_jsonKeys', {'type': 'hidden', 'value': config('_jsonKeys', [])|join(',')})|raw }}
 
         {{ Form.end()|raw }}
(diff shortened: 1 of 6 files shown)


src/Template/Pages/Modules/view.twig
```diff
@@ -21,6 +21,10 @@
             'id': 'form-main',
         })|raw }}
 
+
+            {# Init `_jsonKeys` data #}
+            {{ write_config('_jsonKeys', []) }}
```
xho: Il problema del layout a colonne che si rompe deriva da questa riga: l'istruzione torna un 1 che sta prima delle colonne e spagina. Dante propone per semplicità di assegnare il ritorno a un a var dummy.
I.e. `{% set a = write_config('_jsonKeys', []) %}`

stefanorosanelli: :scream: 
ok, meglio modificare `write_config` to not display anything :wink: 
I'll do that...

stefanorosanelli: fixed

PR #158, opened 2018-08-09
feat: use bedita/web-tools package

batopa:
Code review after the introduction of https://github.com/bedita/web-tools.


src/Shell/CacheShell.php
```diff
@@ -1,40 +0,0 @@
...
-
-namespace App\Shell;
-
-use Cake\Filesystem\Folder;
-use Cake\Shell\CacheShell as BaseCacheShell;
-
-/**
- * Extend `CacheShell::clearAll` to remove Twig compiled files.
- */
-class CacheShell extends BaseCacheShell
```
stefanorosanelli: `Twig` compiled templates are not cleared with `bin/cake cache clearAll`
what about an empty `CacheShell` extending `BEdita\Webtools` one?

PR #161, opened 2018-08-24
feat: trash controller test, use mock for api

didoda:
This provides a refactor in `TrashControllerTest`: use mock api instead of real api client.
Basically bedita-web controller tests should not be aware of php-sdk stuffs... they should just verify that a specific controller method set view data, redirect to a page or have a specific content-type or http status code.
We should remove all "real" api CRUD operations from tests, and use mock api instead.

### DIFF
--- a/tests/TestCase/Controller/TrashControllerTest.php
+++ b/tests/TestCase/Controller/TrashControllerTest.php
@@ -24,6 +24,17 @@ use Cake\TestSuite\TestCase;
 class TrashControllerSample extends TrashController
 {
     /**
+     * Set api client for test (using mock)
+     *
+     * @param \BEdita\SDK\BEditaClient $apiClient The api client
+     * @return void
+     */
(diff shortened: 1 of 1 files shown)


stefanorosanelli:
I'm not convinced that having a mock API is a good idea right now.

* tests are more complex; since we have a low coverage we should try to keep them as simple as possible
* it makes sense to mock an API, theoretically it is also the **right** thing to do but:
  1. some tests where you need to perform more than one API operation become impossible or very difficult
  1. using a real API we also test other parts of the system; we surely have some duplicates test, but it may help to discover some API-related bugs or issues; in other words: `two is megl che one` :smile: 

That said I'd prefer to keep a real API client in our tests here. At least for the moment.

@batopa @fquffio what do you think?

didoda:
Ok. After a talk, we decided to keep full tests (no mock).

PR #175, opened 2018-09-18
Issue/174 relation view fixes

vallo87:
[Issue#174](https://github.com/bedita/web/issues/174)

relation view UX/UI fixes


src/Template/Layout/js/app/components/filter-box.js
```diff
@@ -21,9 +21,9 @@ import { DEFAULT_PAGINATION } from 'app/mixins/paginated-content';
 export default {
     template:
     `
-        <nav class="pagination has-text-size-smallest" :class="pagination.count > 4 && 'show-pagination'">
+        <nav class="pagination has-text-size-smallest" :class="pagination.count > 10 && 'show-pagination'">
```
stefanorosanelli: instead of `10` should this value be the minimum or first from `{{ config('Pagination.sizeAvailable') }}` ?

PR #185, opened 2018-09-26
Issue/150 roles list

vallo87:
[Issue#150](https://github.com/bedita/web/issues/150)

new roles (resource) view and components


src/Template/Layout/js/app/components/relation-view/roles-list-view.js
```diff
@@ -0,0 +1,119 @@
...
+         *
+         * @param {Number} id
+         *
+         * @return {Boolean}
+         */
+        isRelated(id) {
+            return this.relatedObjects.filter((relatedObject) => {
+                return id === relatedObject.id;
+            }).length ? true : false;
+        },
```
didoda: methods `isRemoved` and `isRelated` could be one method, with the objects array as param.
I.e.:
```
containsId(objects, id) 
{
    return objects.filter((obj) => {
        return id === obj.id;
    }).length ? true : false;
}
```

src/Controller/ModulesController.php
```diff
@@ -407,6 +407,34 @@ public function relationData($id, string $relation) : void
...
+     * Json response
+     *
+     * @param string|int $id the object identifier.
+     * @param string $type the resource type name.
+     * @return void
+     */
+    public function resourcesJson($id, string $type) : void
+    {
+        $this->request->allowMethod(['get']);
+        $response = null;
```
didoda: Is Scrutinizer right?

stefanorosanelli: yes, we may remove the above line, `$response` assignment is don later

config/routes.php
```diff
@@ -166,6 +166,11 @@
         ['controller' => 'Modules', 'action' => 'relationshipsJson'],
         ['pass' => ['id', 'relation'], '_name' => 'modules:relationshipsJson']
     );
+    $routes->connect(
+        '/:object_type/view/:id/resourcesJson/:relation',
```
stefanorosanelli: we need to adjust this URL since neither `object_type` nor  object `id` are needed in this context

src/Template/Element/Form/relations.twig
```diff
@@ -7,11 +7,11 @@
 {% if relations %}
 <section class="relations">
 
-    {% for relationName in relations %}
+    {% for relationName in relations if relationName != 'roles' and relationName != 'translations' %}
```
stefanorosanelli: we need to find a better rule to distiguish between `dynamic object relations` and `static object/resources` relations
best solution is to read this info from API

for now we may want to discard only `roles` or have two different arrays from the controller for `dynamic` and `static` relations

src/Template/Layout/js/app/components/relation-view/roles-list-view.js
vallo87: yes, i did it to have a more readable code

PR #186, opened 2018-09-27
WIP: Issue/170 translations

didoda:
This provides:

 - create a new translation
   - view `/:object_type/translation/:id/add`
   - save `/:object_type/translation/save`
 - modify a translation
   - view `/:object_type/translation/:id/:lang`
   - save `/:object_type/translation/save`
 - delete a translation - `/:object_type/translation/delete`

Still TODO (other issue?):
  - handle languages UI (which languages should be allowed in translations? Now: English, Italian)
  - handle languages validation on save (lang combo should avoid that user try to save a new translation, using a lang already used for translations. In this moment system blocks the request with a `400 Bad Request`).

Extra: `css` stuff done by @xho


src/Controller/ModulesController.php
```diff
@@ -173,6 +173,165 @@ public function view($id) : ?Response
         return null;
     }
 
+    /**
+     * Display new translation form.
+     *
+     * @param string|int $id Resource ID.
+     * @return \Cake\Http\Response|null
+     */
+    public function createTranslation($id) : ?Response
```
stefanorosanelli: Why not use a `TranslationsController` instead? 
`ModulesController` is becoming too big...

src/Controller/ModulesController.php
```diff
@@ -173,6 +173,165 @@ public function view($id) : ?Response
...
+        $this->set('object', $object);
+
+
+        return null;
+    }
+
+    /**
+     * View single translation.
+     *
+     * @param string|int $id Resource ID.
```
stefanorosanelli: [minor] I'd say `Object ID` here

src/Controller/ModulesController.php
```diff
@@ -173,6 +173,165 @@ public function view($id) : ?Response
...
+            }
+            if (empty($translation)) {
+                throw new BEditaClientException(sprintf('Translation not found per %s %s and lang %s', $this->objectType, $id, $lang));
+            }
+        } catch (BEditaClientException $e) {
+            // Error! Back to index.
+            $this->log($e, LogLevel::ERROR);
+            $this->Flash->error($e, ['params' => $e->getAttributes()]);
+
+            return $this->redirect(['_name' => 'modules:list', 'object_type' => $this->objectType]);
```
stefanorosanelli: we could probably go back to object `$id` view instead

src/Controller/ModulesController.php
```diff
@@ -173,6 +173,165 @@ public function view($id) : ?Response
...
+     */
+    public function deleteTranslation() : ?Response
+    {
+        $this->request->allowMethod(['post']);
+        $requestData = $this->request->getData();
+        foreach ($requestData as $translation) {
+            if (!empty($translation['id'])) {
+                try {
+                    // remove completely the translation
+                    $this->apiClient->deleteObject($translation['id'], 'translations');
```
stefanorosanelli: we should deprecate `deleteObject` like `saveObject` in order to use `delete`, also valid for resources...

src/Controller/ModulesController.php
```diff
@@ -173,6 +173,165 @@ public function view($id) : ?Response
...
+                } catch (BEditaClientException $e) {
+                    $this->log($e, LogLevel::ERROR);
+                    $this->Flash->error($e, ['params' => $e->getAttributes()]);
+
+                    if (!empty($translation['object_id'])) {
```
stefanorosanelli: shouldn't `object_id` always be present?
I'd handle only this case... or?

src/Template/Pages/Modules/translation.twig
```diff
@@ -0,0 +1,94 @@
...
+                    {% set options = Schema.controlOptions(
+                        'status',
+                        translation.attributes.status|default('draft'),
+                        schema.properties['status']
+                    ) %}
+                    {{ Form.control('status', options)|raw }}
+                </section>
+                <section class="fieldset">
+                    {% set languageLabel = "#{__('Language')} (#{__('original')}: #{object.attributes.lang})" %}
+                    {{ Form.control('lang', {
```
stefanorosanelli: after #190 we can change this part

src/View/Helper/SchemaHelper.php
```diff
@@ -234,4 +238,46 @@ function ($value) {
...
+     */
+    public function translatableFields(array $properties) : array
+    {
+        if (empty($properties)) {
+            return [];
+        }
+
+        $fields = [];
+        foreach ($properties as $name => $property) {
+            if (!empty($property['oneOf'])) {
```
stefanorosanelli: [minor] probably another method to check single `$property` match will make this more readable
[it's a matter of taste... you can leave it like this]

src/Controller/ModulesController.php
didoda: Good point, I was thinking the same... :soon:

src/Controller/ModulesController.php
didoda: On bulk delete also, should we expect `object_id` ( => `[ ['id' => :id, 'object_id' => :object_id], ['id' => :id, 'object_id' => :object_id], ... ]`)? Ok for me

src/View/Helper/SchemaHelper.php
didoda: I would leave `code beautifying` to a "refactor schema helper" issue ;)

stefanorosanelli: I agree

src/Controller/TranslationsController.php
```diff
@@ -0,0 +1,172 @@
...
+    /**
+     * Delete translation(s).
+     *
+     * @return \Cake\Http\Response|null
+     */
+    public function delete() : ?Response
+    {
+        $this->request->allowMethod(['post']);
+        $requestData = $this->request->getData();
+        foreach ($requestData as $translation) {
```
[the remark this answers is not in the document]
didoda: the idea was: handling bulk delete too

src/Controller/TranslationsController.php
```diff
@@ -0,0 +1,172 @@
...
+
+                    // redir to main object view
+                    return $this->redirect(['_name' => 'modules:view', 'object_type' => $this->objectType, 'id' => $translation['object_id']]);
+                }
+            }
+        }
+        $this->Flash->success(__('Translation(s) deleted'));
+
+        // redir to referer
+        return $this->redirect(['_name' => 'modules:view', 'object_type' => $this->objectType, 'id' => $translation['object_id']]);
```
stefanorosanelli: [see scrutinizer note above]

I'd use this rule in case of single delete (`$requestData` count is 1) - otherwise I'll redirect to `$this->object_type` list

PR #188, opened 2018-09-27
Issue/184 translation component

vallo87:
resources template
 -> translation template
~TO-DO links for create new, modify (waiting for issue170 new routes)~


vallo87:
it was due to a missing input field, fixed and ready to go

PR #189, opened 2018-09-28
Modules controller phpunit tests

didoda:
This provides `ModulesControllerTest` class, to test `ModulesController` class.
No mocks used (as decided by BEdita developers team).


tests/TestCase/Controller/ModulesControllerTest.php
```diff
@@ -0,0 +1,772 @@
...
+    public function testDescendants() : void
+    {
+        // Setup controller for test
+        $this->setupController();
+
+        // do controller call
+        $result = $this->controller->descendants();
+
+        // verify response status code and type
+        static::assertTrue(is_array($result));
```
stefanorosanelli: more meaningful test could be if `$result` is empty (on `documents`) or not empty (on `objects`)

didoda: fair enough. coming soon ;)

PR #193, opened 2018-10-05
feat: export csv, extra as columns

didoda:
This provides a more detailed csv export: `extra` data go into `extra-<extrafield>` columns

### DIFF
--- a/src/Controller/ExportController.php
+++ b/src/Controller/ExportController.php
@@ -67,6 +67,9 @@ class ExportController extends AppController
             return;
         }
         $fields = ['id'] + array_keys($response['data']['0']['attributes']);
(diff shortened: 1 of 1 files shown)


src/Controller/ExportController.php
```diff
@@ -86,8 +89,22 @@ public function export() : void
                     }
                 }
             }
+            foreach ($val['attributes']['extra'] as $extrafield => $extraval) {
+                if (is_array($extraval)) {
+                    $row[sprintf('extra-%s', $extrafield)] = json_encode($extraval);
```
stefanorosanelli: I'd use `extra_` as prefix to have consistent naming
Probaly `extra__` even better to avoid possible conflicts

didoda:
another PR did the job

PR #194, opened 2018-10-11
feat: (csv) export all + refactor export selected

didoda:
This provides "Export All" button in modules index page.
It contains MR 193 logic (extra as columns): MR 193 becomes obsolete.


src/Controller/ExportController.php
```diff
@@ -42,32 +43,106 @@ public function beforeFilter(Event $event) : void
...
     public function export() : void
     {
-        if (empty($this->request)) {
-            return;
-        }
-        $this->request->allowMethod(['post']);
-        $ids = $this->request->getData('ids');
-        if (empty($ids)) {
-            return;
+        // check request (allowed methods and required parameters)
```
stefanorosanelli: why not have single `export` method and routing rule? `csvRows` method altready works this way

 * if no `ids` are set all objects - until max limit - are exported
 * if `ids` are set only selected elements

PR #196, opened 2018-10-17
feat: gettext files (master, en, it)

didoda:
This provides translation files (created through `bin/cake gettext update`, updated manually with proper translations); and a basic handling of i18n inside bedita web (=> url to change session lang).

Still to better: managing plugins *.po files + AppController unit tests


src/Template/Element/Form/advanced_properties.twig
```diff
@@ -12,7 +12,7 @@
 
             {% set options = Schema.controlOptions(key, value, schema.properties[key]) %}
 
-            {{ Form.control(key, options)|raw }}
+            {{ Form.control(__(key), options)|raw }}
```
stefanorosanelli: Not sure if we want to translate field names... some yes probably, but we have to think about it

src/Template/Element/Form/relations.twig
```diff
@@ -15,7 +15,7 @@
 
                     <header @click.prevent="toggleVisibility();" class="tab unselectable" v-bind:class="isLoading? 'is-loading-spinner' : ''">
                         <h2>
-                            {{ relationName|humanize|lower }}
+                            {{ __(relationName|humanize)|lower }}
```
stefanorosanelli: I'm quite sure we don't want to translate relation names - we did not in BE3

src/Controller/ModulesController.php
```diff
@@ -553,4 +554,19 @@ public function changeStatus() : ?Response
...
         return $this->redirect(['_name' => 'modules:list', 'object_type' => $this->objectType]);
     }
+
+    /**
+     * Change lang and redir to referer
+     *
+     * @param string $lang The lang code
+     * @return \Cake\Http\Response|null
+     */
+    public function lang($lang) : ?Response
```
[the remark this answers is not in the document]
batopa: Why don't we use cookie for store user lang configuring properly `I18Middleware`?

We could also use `I18n.switchLangUrl` to setup the switch url and avoid this method at all.

The configuration would be something as

```php
Configure::write('I18n', [
    'locales' => [
        'en_US' => 'en',
        'it_IT' => 'it',
    ],
    'default' => 'it',
    'languages' => [
        'en' => 'English',
        'it' => 'Italiano',
    ],
    'cookie' => [
        'name' => 'BEditaWebI18n',
        'create' => true,
    ],
    'switchLangUrl' => '/lang',
]);
```

In this case we should edit `composer.json` to require `"bedita/i18n": "^1.2"`

config/routes.php
```diff
@@ -51,6 +52,9 @@
 
 Router::scope('/', function (RouteBuilder $routes) {
 
+    $locales = array_keys(Configure::read('I18n.languages'));
+    $lang = implode('|', $locales);
```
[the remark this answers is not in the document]
didoda: `$locales` and `$lang` can be removed

src/Template/Element/Form/advanced_properties.twig
didoda: Fair enough. I'd say that we should at least translate `title` and `description`.

src/Controller/ModulesController.php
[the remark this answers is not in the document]
didoda: Good point. I'll try to use cookie

src/Template/Element/Menu/project.twig
```diff
@@ -5,8 +5,15 @@
     {#{% element 'tree' %}#}
     {% element 'Menu/colophon' %}
 
+
     {% if user %}
     <ul class="menu-links">
+        {# Localization link(s) #}
+        {% for code, name in I18n.getLanguages() if code != I18n.getLang() %}
+            <li>
+                {{ Html.link(name, I18n.changeUrlLang(code, config('I18n.switchLangUrl')))|raw }}
```
batopa: Add `, { 'title': name }` as in `menu.twig`

batopa: Nevermind, my mistake

Issue #210, opened 2018-11-16
Messages per auth cases


didoda:
"You are not authorized to access that location" message should appear only when user not logged in try to access a specific path where auth is necessary.

Cases:
 - [ ] login root page (no message)
 - [x] auth failed: "Username or password is incorrect" (already ok)
 - [ ] session expired (message per session expired)
 - [ ] internal url and not logged: "You are not authorized to access that location"

PR #217, opened 2018-11-19
messages in login page

didoda:
This provides #210 

Cases:

 - [x] login root page (no message)
 - [x] auth failed: "Username or password is incorrect" (already ok)
 - [x] session expired OR internal url and not logged: "You are not authorized to access that location"


didoda:
"session expiration" case treated like "internal url and not logged" case (same message: "You are not authorized to access that location")

stefanorosanelli:
Ok, let's use a more meaningful message like:
`You are not logged or your session has expired, please provide login credentials`

src/Controller/AppController.php
```diff
@@ -72,8 +73,17 @@ public function initialize() : void
     public function beforeFilter(Event $event) : ?Response
     {
         $tokens = $this->Auth->user('tokens');
-        if ($tokens) {
+        if (!empty($tokens)) {
             $this->apiClient->setupTokens($tokens);
+        } elseif (!in_array($this->request->url, ['login'])) {
+            $route = ['_name' => 'login'];
+            $redirect = Router::reverse($this->request);
```
stefanorosanelli: did some tests and redirect url is not working as expected for urls like `/documents/view/123`
But it works with:
```php
            $redirect = $this->request->getUri()->getPath();
            if ($redirect !== $this->request->getAttribute('webroot')) {
```

PR #223, opened 2018-11-21
Issue/212 attach and preview

vallo87:
ModuleView -> RelationView: 
shows preview when attaching image to an object

### DIFF
--- a/src/Controller/ModulesController.php
+++ b/src/Controller/ModulesController.php
@@ -459,21 +459,23 @@ class ModulesController extends AppController
                     $available = '/folders';
                     break;
                 default:
-                    $response = $this->apiClient->get($path, $this->request->getQueryParams());// ['page_size' => 1]); // page_size 1: we need just the available
+                    $response = $this->apiClient->get($path, ['page_size' => 1]); // page_size 1: we need just the available
                     $available = $response['links']['available'];
             }
 
             $response = $this->apiClient->get($available, $this->request->getQueryParams());
-        } catch (BEditaClientException $error) {
-            $this->log($error, LogLevel::ERROR);
 
-            $this->set(compact('error'));
-            $this->set('_serialize', ['error']);
+            $this->getThumbsUrls($response);
+        } catch (BEditaClientException $ex) {
+            $this->log($ex, LogLevel::ERROR);
 
-            return;
+            $this->set([
+                'error' => $ex->getMessage(),
+                '_serialize' => ['error'],
+            ]);
+
+            return ;
         }
-
-        $this->getThumbsUrls($response);
 
         $this->set((array)$response);
         $this->set('_serialize', array_keys($response));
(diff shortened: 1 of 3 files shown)


src/Controller/ModulesController.php
```diff
@@ -460,9 +459,10 @@ public function relationshipsJson($id, string $relation) : void
                     $available = '/folders';
                     break;
                 default:
-                    $response = $this->apiClient->get($path, ['page_size' => 1]); // page_size 1: we need just the available link from response
+                    $response = $this->apiClient->get($path, $this->request->getQueryParams());// ['page_size' => 1]); // page_size 1: we need just the available
```
didoda: I would remove the comment, because it's unuseful with new code (it was fine to explain why the filter was `'page_size' => 1`)

src/Controller/ModulesController.php
```diff
@@ -473,10 +473,55 @@ public function relationshipsJson($id, string $relation) : void
...
+     * @return void
+     */
+    protected function getThumbsUrls(array &$response) : void
+    {
+        if (empty($response['data'])) {
+            return;
+        }
+
+        // extract ids of objects
+        $ids = Hash::combine($response, 'data.{n}.id', 'data.{n}.id');
```
didoda: it seemes you need just the ids... `Hash::extract($response, 'data.{n}.id')` could be fine.

src/Controller/ModulesController.php
```diff
@@ -473,10 +473,55 @@ public function relationshipsJson($id, string $relation) : void
...
+
+        if (!count($ids)) {
```
didoda: or..
```
if (empty($ids)) {
    return;
}
```

src/Controller/ModulesController.php
vallo87: i actually wanted to undo this change thanks for spotting it

src/Controller/ModulesController.php
```diff
@@ -477,6 +481,39 @@ public function relationshipsJson($id, string $relation) : void
...
+            return;
+        }
+
+        // extract ids of objects
+        $ids = Hash::extract($response, 'data.{n}.id');
+        if (empty($ids)) {
+            return;
+        }
+
+        $thumbs = '/media/thumbs?ids=' . implode(',', $ids) . '&options[w]=400';
```
stefanorosanelli: * do we try to create thumbs for all object types or only for media types? 
* dont' particularly like the hardcoded `=400`, maybe define a constant in a view or helper class?

tests/TestCase/Controller/ModulesControllerTest.php
```diff
@@ -607,6 +609,104 @@ public function testRelationshipsJson(string $relation, string $objectType) : vo
...
+            ],
+            // correct result
+            'correctResponseMock' => [
+                [ // expected
+                    'data' => [
+                        [
+                            'id' => '43',
+                            'meta' =>
+                                [
+                                    'url' => 'https://s3-eu-west-1.amazonaws.com/be4-media-test/test-thumbs/472f9968-b02e-4253-b65f-0bb95f02b89c-test.png/728b91672d309feb717f12aede127da99b1e47b9.png',
```
stefanorosanelli: [minor] since it's a mock I'd use `example.com` domain, something like `https://media.example.com/thumbs/.....`
to avoid domains actually in use..

src/Controller/ModulesController.php
```diff
@@ -477,6 +481,39 @@ public function relationshipsJson($id, string $relation) : void
...
+     * @return void
+     */
+    public function getThumbsUrls(array &$response) : void
+    {
+        if (empty($response['data'])) {
+            return;
+        }
+
+        // extract ids of objects
+        $ids = Hash::extract($response, 'data.{n}.id');
```
stefanorosanelli: [superminor] add array cast `(array)Hash::extract(...)`

tests/TestCase/Controller/ModulesControllerTest.php
vallo87: ok

src/Controller/ModulesController.php
vallo87: no mention in cake3 doc, what is the reason to be that specific?

stefanorosanelli: scrutinizer suggestion, but we can skip it...

vallo87: i was just curious

PR #226, opened 2018-11-23
beforeFilter and beforeRender return ?Response

didoda:
Refactor to properly use `beforeFilter` and `beforeRender` in controllers.

TODO:
 - verify controllers in plugins
 - human tests

### DIFF
--- a/src/Controller/ErrorController.php
+++ b/src/Controller/ErrorController.php
@@ -15,6 +15,7 @@
 namespace App\Controller;
 
 use Cake\Event\Event;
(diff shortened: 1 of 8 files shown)


src/Controller/ErrorController.php
```diff
@@ -52,12 +54,12 @@ public function beforeFilter(Event $event) : void
      * @return \Cake\Http\Response|null|void
```
stefanorosanelli: `@return \Cake\Http\Response|null`  - remove `void`

PR #229, opened 2018-11-27
change stream in media view

didoda:
This provides the possibility to change a stream, in media view.
When changing a stream, previously associated stream is dropped by system.


src/Controller/Component/ModulesComponent.php
```diff
@@ -222,29 +222,78 @@ public function upload(array &$requestData) : void
...
+            return;
+        }
+
+        $apiClient = ApiClientProvider::getApiClient();
+        $response = $apiClient->get(sprintf('/%s/%s/streams', $requestData['model-type'], $requestData['id']));
+        if (empty($response['data'])) { // no streams for media
+            return;
+        }
+        $streamId = Hash::get($response, 'data.0.id');
+        $apiClient->deleteObject($streamId, 'streams');
```
stefanorosanelli: [memo] we should change method name to `delete` in SDK, since we may also delete resources (like `streams`) that will not go in the trashcan

src/Controller/Component/ModulesComponent.php
```diff
@@ -222,29 +222,78 @@ public function upload(array &$requestData) : void
...
         $filename = $requestData['file']['name'];
         $filepath = $requestData['file']['tmp_name'];
         $headers = ['Content-Type' => $requestData['file']['type']];
         $apiClient = ApiClientProvider::getApiClient();
         $response = $apiClient->upload($filename, $filepath, $headers);
 
-        // create media from stream
+        // assoc stream to media
         $streamId = $response['data']['id'];
+        $requestData['id'] = $this->assocStreamToMedia($streamId, $requestData, $filename);
```
stefanorosanelli: [memo] we probably need something in the API to replace stream without having to remove and add...

src/Template/Element/Form/included.twig
```diff
@@ -84,8 +84,17 @@
         {% endfor %}
         {% endif %}
 
+        <h3>{{ __('Change Stream') }}</h3>
```
stefanorosanelli: I prefer `Change File`

Issue #222, opened 2018-11-21
Can't upload video file


vallo87:
When i try to save a video this error happens

Error message:
```
Invalid form data: file.tmp_name

Error in: /Users/vallo/www/projects/be4-web/src/Controller/Component/ModulesComponent.php, line 218
```

Env:
- master
- feat/207-new-media-remove-upload-button

didoda:
There was a problem in @vallo87 `php.ini`, with unexpected side effect.
He fixed `php.ini` config vars:
```
; Maximum allowed size for uploaded files.
upload_max_filesize = 20M 

; Must be greater than or equal to upload_max_filesize
post_max_size = 20M 
```

Now everything is fine.

stefanorosanelli:
@didoda  Ok, can we have a more meaningful error message?

didoda:
Not really. Php server silently let the post goes with an empty `file.tmp_file` (no warning, no error).
Ipotetically we could do "system check" in bootstrap, to prevent such kind of errors

didoda:
We are going to handle properly upload errors (http://php.net/manual/en/features.file-upload.errors.php).

PR #230, opened 2018-11-27
handle form upload exception

didoda:
This provides #222 handling of form upload errors


src/Core/Exception/UploadException.php
```diff
@@ -0,0 +1,71 @@
...
+     * @see http://php.net/manual/en/features.file-upload.errors.php for details
+     *
+     * @param int $code The php code
+     * @return string
+     */
+    private function codeToMessage(int $code) :string
+    {
+        switch ($code) {
+            case UPLOAD_ERR_INI_SIZE:
+                $message = __('The uploaded file exceeds the upload_max_filesize directive in php.ini');
```
stefanorosanelli: We probably need the actual value via `ini_get('upload_max_filesize')` like:
```php
$message = __('The uploaded file exceeds current max size of {0}', ini_get('upload_max_filesize'));
```

src/Controller/ModulesController.php
```diff
@@ -267,7 +268,7 @@ public function save() : ?Response
 
             // save data
             $response = $this->apiClient->save($this->objectType, $requestData);
-        } catch (BEditaClientException $e) {
+        } catch (\RuntimeException | BEditaClientException | UploadException $e) {
             // Error! Back to object view or index.
             $this->log($e, LogLevel::ERROR);
             $this->Flash->error($e, ['params' => $e->getAttributes()]);
```
stefanorosanelli: From Scrutinizer: ```The method getAttributes() does not exist on RuntimeException. It seems like you code against a sub-type of RuntimeException such as PhpParser\Error or BEdita\SDK\BEditaClientException or Cake\Core\Exception\Exception```
We probably could use `\Cake\Http\Exception\InternalErrorException` instead of `\RuntimeException`

PR #241, opened 2018-12-07
Issue/239 filter box boolean

### LINKED ISSUE #239: filter-box doesn't handle boolean type
add boolean type as radio group to filters

vallo87:
fixes #239 

added boolean type to filter box component


src/Template/Layout/js/app/components/filter-box.js
```diff
@@ -73,6 +73,15 @@ export default {
                         </select>
                     </span>
 
+                    <span v-else-if="filter.type === 'checkbox'">
+                        <label>Both</label>
```
stefanorosanelli: `Any` instead of `Both` ? 

We should also use gettext localization, so something like `{{ __('Yes') }}` - how do we do it here? 
Maybe resolve this in a new PR

vallo87: right, we do it with attributes for now.
we can do it in other ways also. should we create a refactor issue?

stefanorosanelli: yes, create a refactor issue to localize `js` strings via gettext

src/Template/Layout/js/app/components/filter-box.js
```diff
@@ -73,6 +73,15 @@ export default {
...
+                        <input type="radio" v-model="queryFilter.filter[filter.name]" value="">
+                        <label>Yes</label>
+                        <input type="radio" v-model="queryFilter.filter[filter.name]" value="true">
+                        <label>No</label>
+                        <input type="radio" v-model="queryFilter.filter[filter.name]" value="false">
```
stefanorosanelli: why not use a `select` instead of `radio` ?

vallo87: that was @xho request

Issue #242, opened 2018-12-10
Refactor: implement gettext in Vuejs templates


vallo87:

stefanorosanelli:
First step: decide how localized messages may be used in Vue methods and templates.
we should reuse, as much as possible, gettext strings in `Locale/`

PR #246, opened 2018-12-14
gettext in vuejs templates

didoda:
This provides #242, introducing the use of `ttag` and `ttag-cli`.

Usage
=====

Translations in js code
------------------------------

In a js file:
```
import { t } from 'ttag';
let translated = t`something`;
```
In a vue template:
```
template: `
<h1>${t`something`}</h1>
`
```

Updating po files
----------------------

To update po files:
```
$ bin/cake gettext update
```

Dependencies
--------

1. https://github.com/bedita/i18n/issues/15 / https://github.com/bedita/i18n/pull/16


composer.json
```diff
@@ -70,6 +70,12 @@
         "update-dev": [
             "@composer update",
             "@cs-setup"
+        ],
+        "update-locales": [
+            "node_modules/ttag-cli/bin/ttag extract --o src/Locale/en_US/default-js.po --l en src/Template/Layout/js/app",
```
stefanorosanelli: Problem: `it_IT` and `en_US` are not constant predefined locales but the only ones available for now. 

In `bin/cake gettext update` script the rule is to use every locale defined in `src/Locale` we should use same strategy here.

package.json
```diff
@@ -20,7 +20,8 @@
         "bundle-report": "webpack --mode production --report",
         "develop": "webpack --mode development",
         "dev-plugins": "webpack --config ./webpack.config.plugin.js --mode development",
-        "build-plugins": "webpack --config ./webpack.config.plugin.js --mode production"
+        "build-plugins": "webpack --config ./webpack.config.plugin.js --mode production",
+        "ttag": "./node_modules/ttag-cli/bin/ttag po2json src/Locale/en_US/default.po > src/Template/Layout/js/app/locales/en_US.json; ./node_modules/ttag-cli/bin/ttag po2json src/Locale/it_IT/default.po > src/Template/Layout/js/app/locales/it_IT.json"
```
stefanorosanelli: see above commend... same situation here

composer.json
```diff
@@ -70,6 +70,12 @@
...
+            "node_modules/ttag-cli/bin/ttag extract --o src/Locale/it_IT/default-js.po --l it src/Template/Layout/js/app",
```
stefanorosanelli: should these commands be invoked by `bin/cake gettext update` instead??

vallo87: gettext suits only twig templates not ttag (.js) 
therefore there's 3 ways:
1. exec ttag node app from gettext update (very ugly, very quick)
2. modify gettext update so we can parse t`string` key words in .js files (almost useless ,very long)
3. use a quick node script that parse src/Locale and invoke ttag command (not clean, very quick)

i'd go with the third, even though web-tools should be able to do all by itself, it is also true that there's never gonna be a BE frontend app that doesn't depend on node

didoda: I'd suggest using somehow env variables.
I.e.:
```
export $(cat .env | xargs) && ...
```

stefanorosanelli: Suggested solution:
* add a `ttag` extract step from `bin/cake gettext update` - do it only if `ttag` binary and `src/Template/Layout/js/app` folder exists - [why are we using the `Layout` folder anyway??]
* this step should update only the main `master.pot` - merge from master to available .po locales is already done by `gettext update` script - no need to do it twice. Locale .po files are the only i18n sources of truth.
* in the bundle creation copy locale .po files in the bundle - a more advanced solution (to implement later) copy only a subset of each .po files keeping only strings found in js files

vallo87: @stefanorosanelli if i remember @xho moved the js root under Layout sometime ago, nevertheless we already talked about doing changes to the folder structure

vallo87:
i would alse update bedita/i18n dependency to the new version before closing this PR

PR #247, opened 2018-12-19
Issue/244 new media from side panel

### LINKED ISSUE #244: create new media directly from add relation side panel

vallo87:
feat:
for media objects only
when open add relation panel and the related object is a media, it shows a minimalistic form that allows the user to upload a stream, create its related object and stage the new created object, all at once

fixes #244


src/Controller/ModulesController.php
```diff
@@ -302,6 +302,41 @@ public function save() : ?Response
...
+        $this->request->allowMethod(['post']);
+        $requestData = $this->prepareRequest($this->objectType);
+
+        try {
+            // upload file (if available)
+            $this->Modules->upload($requestData);
+
+            // save data
+            $response = $this->apiClient->save($this->objectType, $requestData);
+        } catch (InternalErrorException | BEditaClientException | UploadException $error) {
```
stefanorosanelli: We should probably catch only `BEditaClientException` other exceptions should not be thrown here but handled at higher level..

src/Template/Layout/js/app/components/relation-view/relations-add.js
```diff
@@ -41,18 +41,49 @@ export default {
...
     },
 
+    computed: {
+        /**
+         * return accepted mimetypes
+         *
+         * @return {String} mimetypes
+         */
+        acceptedMimeTypes() {
+            return 'audio/*,video/*,image/*,.pdf';
```
didoda: it would be nice have a config for accepted mime types in `app.php` or elsewhere

stefanorosanelli: The API should tell us about this. Since we currently don't have that informtion I'd prefer to accept everything for now.

PR #256, opened 2019-01-16
feat: new yarn start command to run php server with webpack proxy on it

vallo87:
```
yarn start
```
will launch:
- bin/cake server on port 4545
- webpack web server proxying port 4545

the two separate command are grouped therefore killing the main process will kill them all

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,16 +32,6 @@ export BEDITA_API="{bedita-4-url}"
 export BEDITA_API_KEY="{bedita4-api-key}"
 ```
 
-You are then ready to use the webapp by simply run builtin webserver like this
-
-```bash
-bin/cake server
-```
-
-And then point your browser to `http://localhost:8765/`
-
-Or you can configure your preferred web server like Nginx/Apache and point to `webroot/` as vhost document root.
-
 ## Docker
 
 ### Pull official image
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -82,6 +72,30 @@ Replace `bedita/web:latest` with `be4web-local` (or other chosen name) to lanch
...
+## Custom Dev Environment
+
+If you feel you want to customize ports, vhosts or you have your own php server then you can use these commands
+
+### Cake side App
+
+If you don't have a php enabled web server (apache...) you can use cake builtin server
+
+```bash
+bin/cake server
```
stefanorosanelli: This is not the recommended way to launch web app, not even for development. It only makes sense if you don't have a webserver and want to have a quick look....  

If you are a developer you better use `apache` or `nginx` not cake server.  We should be more clear on that.

package.json
```diff
@@ -16,6 +16,7 @@
     },
     "homepage": "https://github.com/bedita/web#README",
     "scripts": {
+        "start": "(bin/cake server -H 0.0.0.0 -p 4545 & yarn develop --proxy 0.0.0.0:4545)",
```
stefanorosanelli: [see above] ... as a consequence better not add `bin/cake server` script here, since it's not the recommended for development

Issue #258, opened 2019-01-16
Form Helper Error


vallo87:
When opening:
- Model -> relation View
- Model -> object_types View

i get a null error message

PR #259, opened 2019-01-16
#258 fix: model View

vallo87:
check if Model View shows the model of a relation or an object_type correctly 

fixes: #258

### DIFF
--- a/src/Core/Utility/Form.php
+++ b/src/Core/Utility/Form.php
@@ -42,7 +42,7 @@ class Form
      * @param mixed|null $value Property value.
      * @return array
      */
-    public static function control(array $schema, string $type, $value) : array
+    public static function control(array $schema = [], string $type, $value) : array
     {
         if (!in_array($type, Form::CONTROL_TYPES)) {
             return compact('type', 'value');
--- a/src/Core/Utility/Form.php
+++ b/src/Core/Utility/Form.php
@@ -42,7 +42,7 @@ class Form
      * @param mixed|null $value Property value.
      * @return array
      */
-    public static function control(array $schema = [], string $type, $value) : array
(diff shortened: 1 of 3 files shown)


src/Core/Utility/Form.php
```diff
@@ -42,7 +42,7 @@ class Form
      * @param mixed|null $value Property value.
      * @return array
      */
-    public static function control(array $schema, string $type, $value) : array
+    public static function control(array $schema = [], string $type, $value) : array
```
stefanorosanelli: Not a good practice: default arguments should be on the right side of any non-default arguments.
See http://php.net/manual/en/functions.arguments.php `example #5`

src/View/Helper/SchemaHelper.php
```diff
@@ -32,14 +32,16 @@ class SchemaHelper extends Helper
...
-    public function controlOptions($name, $value, $schema) : array
+    public function controlOptions($name, $value, $schema = []) : array
     {
         $options = Form::customControlOptions($name);
         if (!empty($options)) {
             $options['value'] = $value;
 
             return $options;
         }
+        $schema = empty($schema) ? [] : $schema;
```
stefanorosanelli: Not necessary if we use `array $schema = []` as last method argument (add a type)

src/View/Helper/SchemaHelper.php
```diff
@@ -32,7 +32,7 @@ class SchemaHelper extends Helper
      * @param array $schema Object schema array.
      * @return array
      */
-    public function controlOptions($name, $value, $schema = []) : array
+    public function controlOptions($name, $value, $schema) : array
```
stefanorosanelli: [see previous comment] better have:
`controlOptions(string $name, $value, array $schema = [])`

PR #268, opened 2019-02-25
Feat/refactor panel

vallo87:
## Panel refactor

PanelView Component has slots (meaning that every component that has to be used in panel has to go into its template 
```
<panel-view>
    <new-component v-if="props.action === 'action-for-component'" v-bind="props.data" />
</panel-view>
```

### Panel request happens through PanelEvents' event bus

in requester view component use PanelEvents.requestPanel(request);
request has to be an object with an action: String, a callee: VueComponent and a data payload
```
import { PanelEvents } form 'app/components/panel-view';

// to open panel
PanelEvents.requestPanel({
   action: 'editSomething', 
   from: this, 
   data: data,
})
```
### Request Object
```
let request = {
   action: 'editSomething', // used in PanelView to render right component
   from: this, // used in BusEvents to sendback data to right component
   data: data, // payload
}
```

### Listening to panel events

requester would want to listen to save event from panel with
PanelEvents.listen('eventname', requester, callback)
N.B. requester arg is the VueComponent of the caller
```
import { PanelEvents } form 'app/components/panel-view';

PanelEvents.listen('editSomething:save', this, this.editSomethingSaveCallback);
```

### Custom Panel buttons
When using custom buttons in panel' component to send back data to panel requester you need to use PanelEvents.sendBack('eventname', payload);
```
import { PanelEvents } form 'app/components/panel-view';

// panel's rendered component
PanelEvents.sendBack('editSomething:save', payload );
```

## Events

PanelView listen to the following event:
 * @listens panel:request request panel with data
 * @listens panel:send send message through panel bus
 * @listens panel:save save event
 * @listens panel:open open panel (just visibility)
 * @listens panel:hide hide panel (keep rendered component alive)
 * @listens panel:close close panle and flush data

PanelView trigger the following events:
 * @emits panel:requested the panel has been requested
 * @emits panel:saved a save event happened
 * @emits panel:opened an open event happened
 * @emits panel:hidden an hide event happened
 * @emits panel:closed a close event happened


src/Template/Layout/styles/_layout.scss
```diff
@@ -77,10 +81,22 @@ main.layout {
...
+
+    .panel-slot-container {
+        height: 100%;
+    }
+
+    // > footer {
+    //     display: flex;
+    //     justify-content: space-between;
+    //     padding: $gutter;
+    // }
```
didoda: if not necessary, you can get rid of the commented lines

src/Template/Layout/js/app/components/edit-relation-params.js
```diff
@@ -1,40 +1,170 @@
...
+                    </div>
+                </label>
+            </div>
+        </form>
+
+        <footer>
+            <p>
+                <button v-show="isModified"
+                    class="has-background-info has-text-white"
+                    @click.prevent="saveParams()"><: t('Save ') :></button>
```
didoda: Moving view from twig to js, translation references changed.
Update can be done in some steps:

 - launch `bin/cake gettext update`
 - edit properly `*.po` files
 - commit & push

vallo87: fixed, although i noticed that some core properties don't get translated (title, body, lang) we might wanna check into that eventually

didoda: good point: we should put those words (that are not mentioned directly inside templates) in a specific file, `config/i18n_strings.php`

PR #275, opened 2019-02-27
Feat/search in header menu

xho:


xho:
Also feat:

- input focus on open popup
- enter to trigger action
- escape close popup

PR #284, opened 2019-03-08
import async job - unit test failing with post admin async jobs

didoda:
PR opened to investigate on a "strange" error on a unit test that calls api `POST /admin/async_jobs`


didoda:
:latin_cross: it won't be fixed

PR #285, opened 2019-03-18
Issue/282 dynamic component loading

vallo87:
Allow app to import components or vendors (with styles) on demand

### Vue components import vue components (go through vue-loader):
In order to import vue components on demand you just need to write:

```
// vue component/main app 
...
components: {
   ComponentA: () => import(/* webpackChunkName: "whatever-name-i-want" */'path/to/component/component-a'),
}
...
```

### Vue directive import vue components

in bind or inserted method write:
```
import(/* webpackChunkName: "whatever-name-i-want" */'path/to/component/component-a')
                    .then(module => module.default)
                    .then((component) => {
                         ....
                    });
```

## Import vendors

have a dynamic component that import vendors js and style as usual (import Component from '...')
webpack will transpile that to a dynamic import and manage it without further effort)


src/View/Helper/LinkHelper.php
```diff
@@ -177,4 +178,23 @@ public function pluginsJsBundle()
...
+        $jsPath = WWW_ROOT . 'js' . DS . 'vendors' . DS;
+        if (file_exists($jsPath)) {
+            $dir = new Folder($jsPath);
+            $files = $dir->find('.*\.js');
+            foreach ($files as $file) {
+                echo $this->Html->script(sprintf('vendors%s%s', DS, $file));
+            }
+        }
+
+        return;
```
didoda: this statement seems not necessary

src/Template/Layout/js/app/app.js
```diff
@@ -35,20 +22,20 @@ const _vueInstance = new Vue({
...
-        FilterTypeView,
-        EditRelationParams,
         PanelView,
-        UploadFilesView,
-        MainMenu,
-        FlashMessage,
+        ModulesIndex: () => import(/* webpackChunkName: "modules-index" */'app/pages/modules/index'),
+        ModulesView: () => import(/* webpackChunkName: "modules-view" */'app/pages/modules/view'),
+        TrashIndex: () => import(/* webpackChunkName: "trash-index" */'app/pages/trash/index'),
+        TrashView: () => import(/* webpackChunkName: "modules-view" */'app/pages/trash/view'),
```
didoda: should'nt webpackChunkName be `"trash-view"` here?

src/Template/Layout/js/config/locales.js
```diff
@@ -0,0 +1,11 @@
+
+// define locales available to webpack check config/app.php
+// TO-DO dynamic
+const locales = [
+    'en',
+    'it',
+];
+
```
didoda: Locales should be `I18n.locales` set in `config/app.php`, somehow

vallo87: at the moment is the only way to tell moment and flatpikr which locale to use
one solution could be to use a general conf in .env to be read both in webpack and app.default.php

PR #297, opened 2019-04-24
Issue/296 object relationships types

### LINKED ISSUE #296: Review object relations view: relationships types
In a single object view we have 3 possibile relationships:

 * fixed relationships with resources or objects like `streams`, `roles`, `translations`, `parent` - that have a predefined view 
 * dynamic relationships between objects available from `GET /relations` => these should have current dynamic view
 * dynamic relationships with resources that are not in `GET /relations` and that are not in fixed set seen above => in this case we should display only `relation name` and number of related resources in read only mode unless there's a custom view from a plugin for it

vallo87:
fix #296 

- show relations to resource with a different view
- prevent useless multiple fetch for relation schemas


src/Template/Element/Form/relations.twig
```diff
@@ -8,6 +8,7 @@
 <section class="relations">
 
     {% for relationName in relations if relationName != 'roles' and relationName != 'translations' %}
+    {# {{relationsSchema[relationName]}} #}
```
didoda: this line can be removed, I guess

PR #298, opened 2019-04-26
Issue/ckeditor custom button

vallo87:
add custom "add image url" button to ckeditor toolbar
the button prompt the user for an url, then it adds it with the tag


xho:
Closing this Pull request that is obsolete, since we switched to CKEditor 5.
We should analyse and implement image management in the new rich text editor.
@edoardocavazza

PR #318, opened 2019-08-13
feat: #201 save only modified attributes

### LINKED ISSUE #201: View object: save only modified attributes
Only modified attributes should be included in PATCH API call

To prevent unwanted side-effects like NULL fields becoming blank and other concurrency issues.

This should be done in PHP using form data without removing security form settings (CSRF token and the like...)

didoda:
This fixes #201 

Known issue: dates are alwais considered as modified, caused by a different string representation between model data and form data. I.e.:
```
2019-07-02T09:47:01.000Z // form - date-input component
2019-07-02T09:47:01+00:00 // api
```
This issue will be handled in #320


src/Controller/AppController.php
```diff
@@ -168,9 +168,40 @@ protected function prepareRequest($type) : array
...
     }
 
+    /**
+     * Return true if $value1 equals $value2 or both are empty (null|'')
+     *
+     * @param string|array $value1 The first value | field value in model data (db)
+     * @param string|array $value2 The second value | field value from form
+     * @return bool
+     */
+    private function hasChanged($value1, $value2)
```
stefanorosanelli: I'd use ```protected function hasFieldChanged()```

src/Controller/AppController.php
```diff
@@ -168,9 +168,40 @@ protected function prepareRequest($type) : array
...
+    {
+        if (is_bool($value1) && !is_bool($value2)) { // i.e. true / "1"
+            return $value1 !== boolval($value2);
+        }
+        if (empty($value1) && empty($value2)) { // i.e. null / ""
```
stefanorosanelli: I would use a different condition like:
```if (($value1 === null || $value1 === '') && ($value2 === null || $value2 === ''))```

Reason: `0` returns true with `empty()` 
Can we have a `null` value from a field input? :thinking: 
Probably other chenges are necessary.

But we absolutely need some unit tests here

stefanorosanelli (changes requested):
Apart from comments unit tests are absolutely necessary here!

stefanorosanelli (approved):

PR #323, opened 2019-09-19
fix: `create new` option for every media type

stefanorosanelli:
Currently creating a new media object directly from relations panel (`Create new {RelationName}` tab) is possibile only if `media` appears in related types. 
But instead of `media` we can have concrete media types likes `images` or `videos`.

Probably there's a better way to write this.
Media types should be read from API, this could be done via `ModuleComponent` using cached data from `/home` endpoint.


didoda (approved):

src/Template/Layout/js/app/components/relation-view/relations-add.js
```diff
@@ -70,7 +70,12 @@ export default {
          * @return {Boolean}
          */
         isMedia() {
-            return this.relationTypes && this.relationTypes.right.indexOf('media') !== -1;
+            let right = this.relationTypes.right || [];
+            // actual types should be read from API
+            let mediaTypes = ['media', 'audio', 'files', 'images', 'videos'];
+            let intersection = right.filter(x => mediaTypes.includes(x));
```
didoda: minor: `const` could be used instead of `let`

didoda (approved):

Issue #312, opened 2019-05-24
Remote URL in related media objects


stefanorosanelli:
If a new media object is created via `Create New` panel from a relation `add objects` same input form with `Upload local file` or `Embed online content` should be presented.

didoda:
merged

PR #327, opened 2019-09-20
Remote url in related media objects

didoda:
Ref: #312


src/Template/Layout/js/app/components/relation-view/relations-add.js
```diff
@@ -279,6 +299,32 @@ export default {
...
+        /**
+         * Verify if right relation types are fine for url embed.
+         * Now only allowed type is 'video' (and 'media')
+         *
+         * @return {Boolean}
+         */
+        isEmbeddable() {
+            const right = this.relationTypes.right || [];
+
+            return right.includes('video') || right.includes('media');
```
stefanorosanelli: `videos` instead of `video`

stefanorosanelli (changes requested):
Adding a `video` via URL embed in a relation `title` attribute of video is not loaded, instead in single video creation via URL embed `title` is correctly set.

Other fields like `provide *` are instead correctly loaded

didoda:
> Adding a `video` via URL embed in a relation `title` attribute of video is not loaded, instead in single video creation via URL embed `title` is correctly set.
> 
> Other fields like `provide *` are instead correctly loaded

I'll check it... ;)

stefanorosanelli (approved):

Issue #325, opened 2019-09-20
Module filter data in user session


stefanorosanelli:
Module filter and pagination should be saved in user session and used directly as user clicks on module box link (top menu, dashboard or current module).

didoda:
merged

PR #328, opened 2019-09-23
module filter data in user session (#325)

didoda:
This provides #325 

Known issues: 

 - module `objects`: filter by type is stored in session, but not applyed properly when doing a new search
 - index pages: status filter is saved in session and applyed in uri, but not set properly in `status` radio buttons


stefanorosanelli (approved):

didoda:
Replaced by new PR ( #331 ), without bundles and with updated features from `master`

PR #331, opened 2019-10-15
module filter data in user session (#325)

didoda:
This provides #325, and substitutes #328

Known issues:

 - module objects: filter by type is stored in session, but not applyed properly when doing a new search
 - index pages: status filter is saved in session and applyed in uri, but not set properly in status radio buttons
 - active filter highlight can be better ^_^ (now simple bottom border and `Data is filtered` tip)


src/Controller/AppController.php
```diff
@@ -237,4 +237,42 @@ protected function checkRequest(array $options = []) : array
...
+     * Session key: '{$currentModuleName}.filter'
+     * Scenarios:
+     *
+     * Query parameter 'reset=1': remove session key and redirect
+     * Query parameters found: write them on session with proper key ({currentModuleName}.filter)
+     * Session data for session key: build uri from session data and redirect to new uri.
+     *
+     * @return \Cake\Http\Response|null
+     */
+    protected function applySessionFilter()
```
stefanorosanelli: missing return type: `?Response`

src/Controller/ModulesController.php
```diff
@@ -115,7 +115,8 @@ public function index() : ?Response
         // objectTypes schema
         $this->set('schema', $this->getSchemaForIndex($this->objectType));
 
-        return null;
+        // handle filter and query parameters using session
+        return $this->applySessionFilter();
```
stefanorosanelli: This method should be called at the beginning of `index()` not at the end
 * on `null` return data should be loaded
 * in case of redirect no need to load data...

src/Template/Element/FilterBox/filter_box_common.twig
```diff
@@ -1,6 +1,10 @@
 {% set selectBaseClasses = "has-background-gray-700 has-border-gray-700 has-font-weight-light has-text-gray-200 has-text-size-smallest" %}
-
-<div class="filters-container" :class="filterList.length > 1? '' : 'basic-filters-only'">
+{% set containerClass = 'filters-container' %}
+{% set filterActive = _view.request.params['?']|length > 0 %}
```
stefanorosanelli: maybe better:  `set filterActive = (_view.request.getQuery('filter') or _view.request.getQuery('q'))`

otherwise filter is active in case of pagination or other queries...

stefanorosanelli (changes requested):

stefanorosanelli (approved):

Issue #326, opened 2019-09-20
Previous and Next object


stefanorosanelli:
[after #325]
In single object view create a link to previous and next object => links should redirect to previous and next using current module filter.
If an object is last or first no next or prev link should be available.

PR #337, opened 2019-10-25
Object navigation (prev and next links) in modules view

didoda:
Ref: #326

### DIFF
--- a/src/Controller/AppController.php
+++ b/src/Controller/AppController.php
@@ -280,7 +280,15 @@ class AppController extends Controller
     }
 
     /**
-     * Set objectNav array
+     * Set objectNav array and objectNavModule.
+     * Objects can be in different modules:
+     *
+     *  - a document is in "documents" and "objects" index
+     *  - an image is in "images" and "media" index
+     *  - etc.
+     *
+     * The session variable objectNavModule stores the last module index visited;
+     * this is used then in controller view, to obtain the proper object nav (@see \App\Controller\AppController::getObjectNav)
      *
      * @param array $objects The objects to parse to set prev and next data
      * @return void
@@ -301,7 +309,7 @@ class AppController extends Controller
         }
         $session = $this->request->getSession();
         $session->write('objectNav', $objectNav);
-        $session->write('objectTypeNav', $moduleName);
+        $session->write('objectNavModule', $moduleName);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


src/Controller/AppController.php
```diff
@@ -277,4 +278,51 @@ protected function applySessionFilter() : ?Response
...
+            $objectNav[$moduleName][$object['id']] = [
+                'prev' => ($i > 0) ? Hash::get($objects, sprintf('%d.id', $i - 1)) : null,
+                'next' => ($i + 1 < $total) ? Hash::get($objects, sprintf('%d.id', $i + 1)) : null,
+                'index' => $i + 1,
+                'total' => $total,
+                'object_type' => Hash::get($objects, sprintf('%d.object_type', $i)),
+            ];
+        }
+        $session = $this->request->getSession();
+        $session->write('objectNav', $objectNav);
```
stefanorosanelli, line 311: Using `objectNav.$moduleName` as key? We don't need `objectTypeNav` and we could store more lists at the same time (e.g. navigate items in different tabs..)

src/View/Helper/LinkHelper.php
```diff
@@ -233,4 +234,44 @@ protected function findFiles(array $filter, string $type) : array
...
+        $next = '›';
+        $objectType = Hash::get($data, 'object_type', null);
+        if (!empty($data['prev'])) {
+            $prev = $this->Html->link(
+                $prev,
+                [
+                    '_name' => 'modules:view',
+                    'object_type' => $objectType,
+                    'id' => $data['prev'],
+                    'title' => __('prev'),
```
stefanorosanelli: Why add `?title=next` to query string?

stefanorosanelli (changes requested):
Overall ok, good job!
Just a couple of minor suggestions/doubts

src/Controller/AppController.php
didoda: We need the info stored in `objectTypeNav`... it tells the view which was the module index recently used.
The `currentModuleName` is not a good choice... think about a navigation in "objects" index, and then to a view of a "document".  In this case `objectTypeNav = 'objects'` and `currentModuleName = 'documents'`. The object navigation should use `'objects'` (and it does, using `objectTypeNav`)

src/View/Helper/LinkHelper.php
didoda: This is something coming from BEdita3 solution... Let's get rid of it

stefanorosanelli (approved):

PR #352, opened 2019-11-15
feat: on save, recover form data

### LINKED ISSUE #202: View object: submitted form available in case of error
In case of error on single object save we should get previous form and try to fix input data or reload them instead of having to recompile form from scratch (quite annoying....)

didoda:
This fixes #202 using a session key 'failedSaveData'

~Bonus: handle dates in `hasFieldChanged`... so that already saved and not changed dates fields are not saved again.~ [will be done in another issue/PR]

### DIFF
--- a/src/Controller/AppController.php
+++ b/src/Controller/AppController.php
@@ -39,7 +39,7 @@ class AppController extends Controller
     /**
      * {@inheritDoc}
      */
-    public function initialize() : void
+    public function initialize(): void
     {
         parent::initialize();
 
@@ -69,7 +69,7 @@ class AppController extends Controller
     /**
      * {@inheritDoc}
      */
-    public function beforeFilter(Event $event) : ?Response
+    public function beforeFilter(Event $event): ?Response
     {
         $tokens = $this->Auth->user('tokens');
         if (!empty($tokens)) {
@@ -107,7 +107,7 @@ class AppController extends Controller
      *
      * Update session tokens if updated/refreshed by client
      */
-    public function beforeRender(Event $event) : ?Response
+    public function beforeRender(Event $event): ?Response
     {
         if ($this->Auth && $this->Auth->user()) {
             $user = $this->Auth->user();
@@ -132,7 +132,7 @@ class AppController extends Controller
      * @param string $type Object type
      * @return array request data
      */
-    protected function prepareRequest($type) : array
+    protected function prepareRequest($type): array
     {
         // prepare json fields before saving
         $data = (array)$this->request->getData();
@@ -212,7 +212,7 @@ class AppController extends Controller
      * @return array The request data for required parameters, if any
      * @throws Cake\Http\Exception\BadRequestException on empty request or empty data by parameter
      */
-    protected function checkRequest(array $options = []) : array
+    protected function checkRequest(array $options = []): array
     {
         // check request
         if (empty($this->request)) {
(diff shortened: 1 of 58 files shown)


src/Controller/AppController.php
```diff
@@ -332,4 +335,35 @@ protected function getObjectNav($id) : array
...
+        $session->write('failedSaveData', $data);
+    }
+
+    /**
+     * Get data from session by key 'failedSaveData'.
+     * If any, return it and delete session key 'failedSaveData'.
+     *
+     * @return array
+     */
+    protected function getDataFromFailedSave() : array
```
stefanorosanelli: Id' use a different method (maybe in `ModulesComponent` ?) 

```php
function updateFromFailedSave(array &$object) {
    // check session data for object id => use `failedSave.{type}.{id}` as key
    // if data exist for {type} and {id} and `__timestamp` (see comment above) not too old (<= 5 minutes ?)
    //  => merge with $object['attributes']
    // remove session data anyway 
   $session->delete('failedSave.{type}.{id}');
}

File: src/Controller/AppController.php [hunk 2]
```diff
@@ -332,4 +335,35 @@ protected function getObjectNav($id) : array
...
+
+    /**
+     * Set session data for key 'failedSaveData'.
+     *
+     * @param array $data The data to store into 'failedSaveData'.
+     * @return void
+     */
+    protected function setDataFromFailedSave($data)
+    {
+        $session = $this->request->getSession();
```
stefanorosanelli: I'd use `failedSave.{type}.{id}` as key and add a `__timestamp` attribute with current timestamp

src/Controller/AppController.php
```diff
@@ -192,14 +192,17 @@ protected function prepareRequest($type) : array
...
     protected function hasFieldChanged($value1, $value2)
     {
-        if (($value1 === null || $value1 === '') && ($value2 === null || $value2 === '')) {
+        if (($value1 === $value2) || (($value1 === null || $value1 === '') && ($value2 === null || $value2 === ''))) {
             return false;
         }
         if (is_bool($value1) && !is_bool($value2)) { // i.e. true / "1"
             return $value1 !== boolval($value2);
         }
+        if (is_string($value1) && is_string($value2) && strtotime($value1) && strtotime($value2)) {
```
stefanorosanelli: This is to make datetime/timestamp comparisons?
Not sure of possible side effects
Can we do it in another PR?
It could be probably safer to have a property type info from schema for a better check...

stefanorosanelli (changes requested):

src/Controller/AppController.php
```diff
@@ -332,4 +332,22 @@ protected function getObjectNav($id): array
...
     }
+
+    /**
+     * Set session data for `failedSave.{type}.{id}` and `failedSave.{type}.{id}__timestamp`.
+     *
+     * @param string $type The object type.
+     * @param array $data The data to store into session.
+     * @return void
+     */
+    protected function setDataFromFailedSave($type, $data): void
```
stefanorosanelli: move also to `ModulesComponent` ? 
is probably convenient to have both methods writing and reading this data in session in the same class

src/Controller/ModulesController.php
```diff
@@ -292,13 +296,18 @@ public function save(): ?Response
             // upload file (if available)
             $this->Modules->upload($requestData);
 
+            throw new BEditaClientException('DEBUG');
```
stefanorosanelli: you probably forgot this one :smile_cat:

stefanorosanelli (changes requested):

stefanorosanelli (approved):

PR #359, opened 2019-11-28
Safe login redirects

### LINKED ISSUE #357: Handle requests to save actions when session is expired
Trying to save an object when session is expired:

1. first redirects the user to the login page;
2. than there is another redirect to the original save action after a successful login.

This throws errors since information is lost in redirects.

- As a quick solution I would suggest to avoid redirects to save actions after login.

- For a more articulate solution I would keep the information about the session (if it is still live) in the javascript environment, in order to show a blocking popup containing an AJAX login form when the session is expired (and force a new login before proceeding).

didoda:
This fixes #357 

When request is not get, after login redirect points to '/'.
When request is get and is webroot, after login redirect points to '/'.
Otherwise, normal redirects to referer.

### DIFF
--- a/src/Controller/AppController.php
+++ b/src/Controller/AppController.php
@@ -88,7 +88,7 @@ class AppController extends Controller
     /**
      * Return route array for login redirect.
      * When request is not a get, return route without redirect.
-     * When request uri path has attribute webroot, return route without redirect.
+     * When request uri path equals request attribute webroot (the app 'webroot'), return route without redirect.
      * Return route with redirect, otherwise.
      *
      * @return array
@@ -102,7 +102,7 @@ class AppController extends Controller
             return $route;
         }
 
-        // if redirect is webroot, return route without redirect.
+        // if redirect is app webroot, return route without redirect.
         $redirect = $this->request->getUri()->getPath();
         if ($redirect === $this->request->getAttribute('webroot')) {
             return $route;
--- a/src/Controller/AppController.php
+++ b/src/Controller/AppController.php
@@ -93,7 +93,7 @@ class AppController extends Controller
      *
      * @return array
      */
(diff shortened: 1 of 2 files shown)


src/Controller/AppController.php
```diff
@@ -89,6 +85,32 @@ public function beforeFilter(Event $event): ?Response
...
+    {
+        $route = ['_name' => 'login'];
+
+        // if request is not a get, return route without redirect.
+        if (!$this->request->is('get')) {
+            return $route;
+        }
+
+        // if redirect is webroot, return route without redirect.
+        $redirect = $this->request->getUri()->getPath();
```
stefanorosanelli, line 106: Usually requests lie `JS/CSS` or other resources in webroot should not be passed to the PHP app, we should ignore this case.
Request URL could be wrong, but we can do nothing about that...

stefanorosanelli (changes requested):

src/Controller/AppController.php
didoda: > Usually requests lie `JS/CSS` or other resources in webroot should not be passed to the PHP app, we should ignore this case.
> Request URL could be wrong, but we can do nothing about that...

A misunderstanding: this `webroot` is not js/css folder; it's the app webroot string.
Maybe phpdoc confused. I updated it.

didoda: > Usually requests lie `JS/CSS` or other resources in webroot should not be passed to the PHP app, we should ignore this case.
> Request URL could be wrong, but we can do nothing about that...

A misunderstanding: this `webroot` is not js/css folder; it's the app webroot string.
Maybe phpdoc confused. I updated it.

stefanorosanelli (approved):

PR #369, opened 2019-12-05
Export controller tests

### LINKED ISSUE #365: Add ExportController tests
`ExportController` has no unit tests, hence no coverage

didoda:
This resolves #365

### DIFF
--- a/src/Controller/ExportController.php
+++ b/src/Controller/ExportController.php
@@ -24,7 +24,7 @@ class ExportController extends AppController
     /**
      * Export data in csv format
      *
-     * @return Cake\Http\Response
+     * @return \Cake\Http\Response
      */
     public function export(): Response
     {
@@ -125,7 +125,7 @@ class ExportController extends AppController
      *
      * @param array $rows The rows data
      * @param string $objectType The object type
-     * @return Cake\Http\Response
+     * @return \Cake\Http\Response
      */
     private function csv(array $rows, string $objectType): Response
     {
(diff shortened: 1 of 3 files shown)


src/Controller/ExportController.php
```diff
@@ -128,9 +128,43 @@ private function fillDataFromResponse(array &$data, array $response): array
...
+    /**
+     * Output csv file.
+     *
+     * @param string $filename The file name.
+     * @param string $csv The csv string data.
+     * @return void
+     */
+    protected function outputCsv($filename, $csv): void
+    {
+        header("Content-type: text/csv;charset=UTF-8");
```
stefanorosanelli: We should probably use https://book.cakephp.org/3/en/controllers/request-response.html#sending-files 
With some additional headers probably... 
This way we should avoid to call `exit` and create mock

stefanorosanelli (changes requested):

stefanorosanelli (approved):

tests/TestCase/Controller/ExportControllerTest.php
```diff
@@ -0,0 +1,564 @@
...
+     */
+    public function testCsvRows($input, $expected): void
+    {
+        // mock api getObjects.
+        $apiClient = $this->getMockBuilder(BEditaClient::class)
+            ->setConstructorArgs(['https://api.example.org'])
+            ->getMock();
+        $apiClient->method('getObjects')
+            ->willReturn($expected['response']);
+        ApiClientProvider::setApiClient($apiClient);
```
stefanorosanelli, line 228: is this necessary?
I suspect this causes other tests to fail, see https://travis-ci.org/bedita/manager/jobs/621166307 
Some mocked API client is then used in other test cases...

didoda: resetting apiclient in tearDown... does the job

Issue #397, opened 2020-05-15
Improve folders navigation


stefanorosanelli:
In `Folders` list view and `Folders` tab in object view we need some changes to:

- display all roots, and dynamically load subfolders upon user action - see image below

![new-folders](https://user-images.githubusercontent.com/2122280/82043683-51d9f000-96ac-11ea-9439-bd8965b5dbb0.jpg)

- at every level the number of folders is limited by pagination (20 or 50 items for instance), with a `More...` link or button more folders are loaded if necessary

- in case of `small` trees it would be nice to have all levels open and listed without user action

stefanorosanelli:
Closed in #398 and in #400

PR #398, opened 2020-05-27
WIP: Tree folders view

edoardocavazza:
This is a tracking PR that introduces a new tree folders navigation as described in #397.

- [x] Dynamically load children
- [x] Styling
- [x] Improve routing
- [x] Handle children pagination
- [x] Restore parents in tree
- [x] Save object position
- [x] Save folder position
- [x] Add tests

### DIFF
--- a/config/routes.php
+++ b/config/routes.php
@@ -202,11 +202,6 @@ Router::scope('/', function (RouteBuilder $routes) {
     $routes->connect(
         '/:object_type/view/:id/treeJson',
         ['controller' => 'Modules', 'action' => 'treeJson'],
-        ['pass' => ['id', 'root'], '_name' => 'modules:treeRootJson']
-    );
-    $routes->connect(
-        '/:object_type/view/:id/treeJson/:root',
-        ['controller' => 'Modules', 'action' => 'treeJson'],
         ['pass' => ['id', 'root'], '_name' => 'modules:treeJson']
     );
     $routes->connect(
--- a/config/routes.php
+++ b/config/routes.php
@@ -200,9 +200,14 @@ Router::scope('/', function (RouteBuilder $routes) {
         ['pass' => ['id', 'relation'], '_name' => 'modules:resourcesJson']
     );
     $routes->connect(
-        '/:object_type/view/:id/treeJson/:relation',
+        '/:object_type/view/:id/treeJson',
         ['controller' => 'Modules', 'action' => 'treeJson'],
-        ['pass' => ['id', 'relation'], '_name' => 'modules:treeJson']
+        ['pass' => ['id', 'root'], '_name' => 'modules:treeRootJson']
+    );
+    $routes->connect(
+        '/:object_type/view/:id/treeJson/:root',
+        ['controller' => 'Modules', 'action' => 'treeJson'],
+        ['pass' => ['id', 'root'], '_name' => 'modules:treeJson']
     );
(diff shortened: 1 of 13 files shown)


src/Controller/ModulesController.php
```diff
@@ -458,6 +462,44 @@ public function relatedJson($id, string $relation): void
...
+                $query['filter']['roots'] = '';
+                $response = $this->apiClient->getObjects('folders', $query);
+            } else {
+                if (empty($query['filter'])) {
+                    $query['filter'] = [];
+                }
+                $query['filter']['parent'] = $root;
+                $response = $this->apiClient->getObjects('folders', $query);
+            }
+        } catch (BEditaClientException $error) {
```
didoda, line 489: you could simplify code [lines 477-489]; i.e.:

```
$key = (empty($root)) ? 'roots' : 'parent';
$val = (empty($root)) ? '' : $root;
if (empty($query['filter'])) {
    $query['filter'] = [];
}
$query['filter'][$key] = $val;
$response = $this->apiClient->getObjects('folders', $query);
```

src/Controller/ModulesController.php
```diff
@@ -458,6 +462,38 @@ public function relatedJson($id, string $relation): void
...
+     * @param string|int $root the tree root identifier.
+     * @return void
+     */
+    public function treeJson($id, string $root = null): void
+    {
+        $this->request->allowMethod(['get']);
+        $query = $this->Modules->prepareQuery($this->request->getQueryParams());
+        try {
+            $key = (empty($root)) ? 'roots' : 'parent';
+            $val = (empty($root)) ? 'roots' : $root;
```
didoda: `$val = (empty($root)) ? '' : $root;`

config/routes.php
```diff
@@ -200,9 +200,14 @@
         ['pass' => ['id', 'relation'], '_name' => 'modules:resourcesJson']
     );
     $routes->connect(
-        '/:object_type/view/:id/treeJson/:relation',
+        '/:object_type/view/:id/treeJson',
```
stefanorosanelli, line 203: you could keep only the first rule and check `root` as optional query string in `treeJson` method

[depending on how you generate routes...]

src/Controller/AppController.php
```diff
@@ -182,7 +182,11 @@ protected function prepareRequest($type): array
                 $id = $data['id'];
 
                 foreach ($relationData as $method => $ids) {
-                    $relatedIds = json_decode($ids, true);
+                    $relatedIds = is_string($ids) ? json_decode($ids, true) : array_map(
```
stefanorosanelli: In order to ease debugging we avoid ternary `?` operator usually... 

```php
     if (is_string($ds)) {
           $relatedIds = is_string($ids);
     } else { 
           $relatedIds = array_map(...);
     }
```

but you can keep it like this

stefanorosanelli (approved):

config/routes.php
```diff
@@ -200,9 +200,9 @@
...
         ['controller' => 'Modules', 'action' => 'treeJson'],
-        ['pass' => ['id', 'relation'], '_name' => 'modules:treeJson']
+        ['pass' => ['id', 'root'], '_name' => 'modules:treeJson']
```
stefanorosanelli: we can remove `root' from `pass` array now since is not passed as argument to `treeJson` method anymore

PR #400, opened 2020-06-01
Navigate folders using tree view

edoardocavazza:
This PR completes #397 by using the tree view as default view for folders.

### DIFF
--- a/src/Controller/AppController.php
+++ b/src/Controller/AppController.php
@@ -12,7 +12,6 @@
  */
 namespace App\Controller;
 
-use BEdita\SDK\BEditaClientException;
 use BEdita\WebTools\ApiClientProvider;
 use Cake\Controller\Controller;
 use Cake\Core\Configure;
@@ -373,41 +372,5 @@ class AppController extends Controller
         $objectNavModule = (string)$session->read('objectNavModule');
 
         return (array)Hash::get($objectNav, sprintf('%s.%s', $objectNavModule, $id), []);
-    }
-
-    /**
-     * Load tree data.
-     *
-     * @return void
-     */
-    public function treeJson(): void
-    {
-        $this->request->allowMethod(['get']);
-        $query = $this->Modules->prepareQuery($this->request->getQueryParams());
-        $root = Hash::get($query, 'root');
-        $full = Hash::get($query, 'full');
-        unset($query['full']);
-        unset($query['root']);
-        try {
-            if (empty($full)) {
-                $key = (empty($root)) ? 'roots' : 'parent';
-                $val = (empty($root)) ? '' : $root;
-                if (empty($query['filter'])) {
-                    $query['filter'] = [];
-                }
-                $query['filter'][$key] = $val;
-            }
-            $response = $this->apiClient->getObjects('folders', $query);
-        } catch (BEditaClientException $error) {
-            $this->log($error, 'error');
-
-            $this->set(compact('error'));
-            $this->set('_serialize', ['error']);
-
-            return;
-        }
-
-        $this->set((array)$response);
-        $this->set('_serialize', array_keys($response));
     }
 }
--- a/src/Controller/AppController.php
+++ b/src/Controller/AppController.php
@@ -12,6 +12,7 @@
  */
 namespace App\Controller;
 
+use BEdita\SDK\BEditaClientException;
 use BEdita\WebTools\ApiClientProvider;
 use Cake\Controller\Controller;
 use Cake\Core\Configure;
@@ -398,7 +399,7 @@ class AppController extends Controller
             }
             $response = $this->apiClient->getObjects('folders', $query);
         } catch (BEditaClientException $error) {
-            $this->log($error, LogLevel::ERROR);
+            $this->log($error, 'error');
 
             $this->set(compact('error'));
             $this->set('_serialize', ['error']);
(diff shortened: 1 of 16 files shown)


tests/TestCase/Controller/AppControllerTest.php
```diff
@@ -848,4 +848,32 @@ public function testGetObjectNav(string $moduleName, array $objects): void
         // verify objectNavModule
         static::assertEquals($session->read('objectNavModule'), $moduleName);
     }
+
+    /**
+     * Test `treeJson` method
+     *
+     * @param array $expected The expected data
```
didoda: la funzione non ha param, questa riga di phpdoc si può rimuovere

src/Template/Pages/Modules/index.twig
```diff
@@ -1,155 +1,159 @@
...
+                                    {% set options = options|merge({ 'name': 'attributes[' ~ key ~ ']' }) %}
+
+                                    {% if options.type == 'checkbox' %}
+                                        {# custom radio input for type checkbox #}
+                                        {% set radio =  {
+                                            "type":"radio",
+                                            "options":
+                                                [
+                                                    { "value": 1, "text": "Yes" },
+                                                    { "value": 0, "text": "No" }
```
didoda: minor: translage "Yes" and "No": `"{{ __('Yes') }}"`, `"{{ __('No') }}"`

didoda (approved):

src/Controller/AppController.php
```diff
@@ -373,4 +373,40 @@ protected function getObjectNav($id): array
...
+            if (empty($full)) {
+                $key = (empty($root)) ? 'roots' : 'parent';
+                $val = (empty($root)) ? '' : $root;
+                if (empty($query['filter'])) {
+                    $query['filter'] = [];
+                }
+                $query['filter'][$key] = $val;
+            }
+            $response = $this->apiClient->getObjects('folders', $query);
+        } catch (BEditaClientException $error) {
```
stefanorosanelli: missing import => `use BEdita\SDK\BEditaClientException;`

src/Controller/AppController.php
```diff
@@ -373,4 +373,40 @@ protected function getObjectNav($id): array
...
+            $this->log($error, LogLevel::ERROR);
```
stefanorosanelli: also missing import for `LogLevel` 

but we can safely use ` $this->log($error, 'error');` avoiding import

src/Controller/ModulesController.php
```diff
@@ -108,6 +110,7 @@ public function index(): ?Response
         $this->set('meta', (array)$response['meta']);
         $this->set('links', (array)$response['links']);
         $this->set('types', ['right' => $this->descendants()]);
+        $this->set('query', $query);
```
stefanorosanelli: no need to pass `query`

in a Twig view you can use

```
{% set query = _view.getRequest().getQuerytParams() %}

{% set filter = _view.getRequest().getQueryt('filter') %}
```

src/Controller/AppController.php
```diff
@@ -373,4 +373,40 @@ protected function getObjectNav($id): array
 
         return (array)Hash::get($objectNav, sprintf('%s.%s', $objectNavModule, $id), []);
     }
+
+    /**
+     * Load tree data.
+     *
+     * @return void
+     */
+    public function treeJson(): void
```
stefanorosanelli: If this method is used only on `/treeJson` route we could use a new `TreesController` class for it

If instead it may be used by other controllers we can leave it here

stefanorosanelli (changes requested):

src/Controller/AppController.php
edoardocavazza: ¯\_(ツ)_/¯

As @didoda suggested, if we want to reuse this logic we could also use a Component and still have a TreeController.

src/Controller/TreeController.php
```diff
@@ -0,0 +1,58 @@
...
+        $query = $this->Modules->prepareQuery($this->request->getQueryParams());
+        $root = Hash::get($query, 'root');
+        $full = Hash::get($query, 'full');
+        unset($query['full']);
+        unset($query['root']);
+        try {
+            if (empty($full)) {
+                $key = (empty($root)) ? 'roots' : 'parent';
+                $val = (empty($root)) ? '' : $root;
+                if (empty($query['filter'])) {
```
stefanorosanelli: since we are setting a filter with `$query['filter'][$key] = $val` adding an empty filter with `$query['filter']=[]` seems unnecessary... or am I missing something?

tests/TestCase/Controller/TreeControllerTest.php
```diff
@@ -0,0 +1,75 @@
...
+                'REQUEST_METHOD' => 'GET',
+            ],
+            'get' => [],
+            'params' => [
+                'object_type' => 'folders',
+            ],
+        ]);
+
+        // do controller call
+        $this->AppController->treeJson();
```
stefanorosanelli: it's `$this->Tree->...` probably? also below

stefanorosanelli (changes requested):

src/Controller/TreeController.php
edoardocavazza: 👍

tests/TestCase/Controller/TreeControllerTest.php
edoardocavazza: 👍

stefanorosanelli (approved):

Issue #403, opened 2020-06-08
Handle date ranges in view (calendar) and index page (column date_ranges)


didoda:
In add/edit page, in calendar box, it should be possible "handling" all date_ranges couples (now only first couple publish_start, publish_end).

handling: add / edit / remove

In index page, when column date_ranges is viewable, all date ranges should be visible (now only first couple publish_start, publish_end).

stefanorosanelli:
Closed in #405

PR #405, opened 2020-06-09
#403 handle date ranges

didoda:
This provides #403 
Handle date ranges: add, edit, remove


src/Template/Element/Form/date_ranges.twig
```diff
@@ -0,0 +1,15 @@
+<date-ranges-list inline-template data-items={{ dateRanges|json_encode }}>
```
stefanorosanelli: Since this is not part of a form we may wanto to use another path like `Element/Index/date_ranges.twig`

src/Template/Layout/js/app/components/date-ranges-list/date-ranges-list.js
```diff
@@ -0,0 +1,37 @@
...
+        this.dateRanges = JSON.parse(this.$el.dataset.items);
+        this.toggle = false;
+    },
+
+    /**
+     * @inheritDoc
+     */
+    methods: {
+        format(date) {
+            return flatpickr.formatDate(new Date(date), 'Y-m-d h:i K');
```
stefanorosanelli: do we need flatpickr to simply display a date string?

[flatpickr could be replaced or removed in a future release]

src/Template/Pages/Modules/index.twig
```diff
@@ -51,7 +51,13 @@
                             <div class="thumb-cell">{% element 'Modules/thumb' { 'object': object } %}</div>
 
                             {% for prop in properties %}
-                                <div class="{{ prop }}-cell" untitled-label="{{ __('Untitled') }}">{{ Schema.format(object.attributes[prop], schema.properties[prop]) }}</div>
+                                <div class="{{ prop }}-cell" untitled-label="{{ __('Untitled') }}">
+                                    {% if (prop == 'date_ranges') %}
```
stefanorosanelli: We need something less hard-coded in order to allow use of `element` to display some attributes in a list.
Just to avoid having a long `{% if ... %} {% elsefi %}..` chain in the future.

For example via a new `Schema.listItem()` method (or similar name) to load a custom element or call `Schema.format()`

stefanorosanelli (changes requested):
Some minor adjustements:

* if no date ranges are set I'd expect to see `Start Date` - `End Date` empty input fields, but no input fields are displayed
* if you remove the last `Start Date/End Date` pair and save object those dates are still present => removing last `Start Date/End Date` I expect:
   - see date input fields (present, but empty) 
   - upon save no date should be present

stefanorosanelli (approved):

PR #406, opened 2020-06-10
Feat/object categories management

DarioSacco:
WIP PR to add categories tab in object view pages.

### DIFF
--- a/src/Controller/ModulesController.php
+++ b/src/Controller/ModulesController.php
@@ -184,32 +184,7 @@ class ModulesController extends AppController
         $revision = Hash::get($response, 'meta.schema.' . $this->objectType . '.revision', null);
         $schema = $this->Schema->getSchema($this->objectType, $revision);
 
-        // REMOVE THIS
-        $schema['categories'] = [
-            [
-                'name' => 'geografia e viaggi',
-                'label' => 'Geografia e Viaggi',
-            ],
-            [
-                'name' => 'letteratura',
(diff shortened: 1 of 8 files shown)


src/Controller/ModulesController.php
```diff
@@ -184,16 +184,41 @@ public function view($id): ?Response
...
+            ],
+            [
+                'name' => 'sport e giochi',
+                'label' => 'Sport e giochi',
+            ],
+        ];
+
         $object = $response['data'];
 
+        // set checked state of categories
```
stefanorosanelli: Can `$category['checked']` set be performed in `Helper` or `Form` utility class? 
It seems very `view` specific... 

ModulesController::view() is already doing too much stuff...

src/Controller/ModulesController.php
```diff
@@ -184,16 +184,41 @@ public function view($id): ?Response
...
+                $category['checked'] = in_array($category['name'], $names);
+            }
+        }
+
         // setup `currentAttributes` and recover failure data from session.
         $this->Modules->setupAttributes($object);
 
         $included = (!empty($response['included'])) ? $response['included'] : [];
-        $this->set(compact('object', 'included', 'schema'));
+        $this->set(compact('object', 'included', 'schema', 'categories'));
```
stefanorosanelli: do we need to set `categories` when we have it in `schema['categories']` ?

stefanorosanelli (changes requested):

PR #413, opened 2020-06-17
fix: remove missing favicons from meta

didoda:
Remove from meta favicons links (not in webroot folder).
This avoids a 404 error on loading favicons.

### DIFF
--- a/src/Template/Element/meta.twig
+++ b/src/Template/Element/meta.twig
@@ -2,8 +2,6 @@
 <meta name="description" content="">
 <meta name="author" content="">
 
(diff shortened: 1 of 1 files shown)


stefanorosanelli (approved):

src/Template/Element/meta.twig
```diff
@@ -3,8 +3,6 @@
 <meta name="author" content="">
 
 <link rel="apple-touch-icon" sizes="180x180" href="/apple-touch-icon.png">
```
xho: Not even this icon exists in `webroot`, nobody noticed the error because it is used only by apple mobile devices. I would remove this line too.

Issue #416, opened 2020-06-23
Datepicker date input, value not visible in "other properties"


didoda:
Datepicker input fields do not show date selected if form control is in "other properties" (or in an hidden section)
For example, in users view, "birthdate" is not visible as expected.

xho:
Solved in https://github.com/bedita/manager/pull/420

PR #420, opened 2020-06-29
fix: datepicker, also handle attrs on componentUpdated

xho:
Solves https://github.com/bedita/manager/issues/416


xho:
Solves https://github.com/bedita/manager/issues/416

stefanorosanelli (approved):
Tested localy, it works now!

PR #421, opened 2020-06-30
Date ranges filter

stefanorosanelli:
This PR adds a first version of `date_ranges` based filter

### DIFF
--- a/src/Template/Element/FilterBox/filter_box_common.twig
+++ b/src/Template/Element/FilterBox/filter_box_common.twig
@@ -75,7 +75,7 @@
             </label>
         </span>
 
-        <span v-if="filter.date && (filter.name !== 'date_ranges')">
+        <span v-else-if="filter.date">
             <label>{{ __('From') }}
             <input-dynamic-attributes :value.sync="queryFilter.filter[filter.name]['gte']" :attrs="filter" />
             </label>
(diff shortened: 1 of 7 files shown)


src/Template/Element/FilterBox/filter_box_common.twig
```diff
@@ -66,12 +66,21 @@
...
+        <span v-if="filter.name === 'date_ranges'">
+            <label>{{ __('From date') }}
+            <input-dynamic-attributes :value.sync="queryFilter.filter['date_ranges']['from_date']" :attrs="filter" />
+            </label>
+            <label>{{ __('To date') }}
+            <input-dynamic-attributes :value.sync="queryFilter.filter['date_ranges']['to_date']" :attrs="filter" />
+            </label>
+        </span>
+
+        <span v-if="filter.date && (filter.name !== 'date_ranges')">
```
DarioSacco: maybe this can be shortened with `v-else-if="filter.date"`

stefanorosanelli: you're right, but we should probably apply `v-else-if` to previous  `<span>` as well and add a final `v-else` ?

DarioSacco: yes `v-else-if` is ok on the second span checking `filter.type`'s value. final `v-else` is ok if the value of `filter.date` isn't relevant to show its content

stefanorosanelli: ok, have look now

Issue #435, opened 2020-07-21
Relation - edit params - oneOf properties


didoda:
# BEdita4 relations params

In BEdita4, a relation params (a record of `object_relations.params`) contains structured data, defined by its `properties` (structure in `relations.params` for the specified relation).
Each property has its own json api format.
An example of params definition:
```
{
    "definitions":{}
    "$schema":"http:\/\/json-schema.org\/draft-06\/schema#",
    "type":"object",
    "properties":{
        "status":{
            "type":"string",
            "description":"The status",
            "enum":["pending","failed","accepted","rejected","unsubscribed"]
        },
        "subscribed_date":{
            "oneOf":[
                {"type":"null"},
                {"type":"string","format":"date-time"}
            ],
            "description":"The subscribed date"
        },
        "visibility":{
            "type":"string",
            "description":"The visibility of the status",
            "enum":["show","hide"]
        }
    }
}
```
# BE4 manager: edit params

In modules view, relation box: when clicking on "edit params" button, manager opens a side panel with form data, representing `:relation.params`. Side panel shows only fields whose type is: `boolean`, `string`, `number`, `integer`.

Side panel could show all json api format fields (i.e. `oneOf`, `anyOf`, `allOf`, `not`; ref: https://json-schema.org/understanding-json-schema/reference/combining.html). By now, BEdita4 allows only a subset of json schema: panel should render `oneOf` properties too.

didoda:
Solved by #438

PR #436, opened 2020-07-21
feat: property oneOf in relation params

didoda:
This provides #435


didoda:
Obselete. Better PR: #437 437

PR #439, opened 2020-07-26
Issue/432 remove relations staging area

### LINKED ISSUE #432: Remove staging area for new relations

xho:


xho:
No particular problems emerged when the staging area was removed, but there is an UX issue to solve later. When you add new objects, they are appended at the end of the page that is currently displayed. In order to keep the UI consistent they should keep the relative position on save. So, still to be done: i think that the priority of all the related objects should be saved.

qwerg:
Ok

qwerg (approved):

PR #443, opened 2020-09-01
Handle relations, categories and date ranges in new object form

stefanorosanelli:
[bedita/bedita#1710 is recommended] 

This PR enables use of related objects, folders, categories/tags and calendar date ranges on newly created objects.

New route for new objects form has been created `/:object_type/view/new` to have consistent rules on Ajax calls like `/:object_type/view/:id/relatedJson/:relation`

`new` URL path final part could probably be changed to other strings like `_new` or `0`

### DIFF
--- a/src/Controller/ModulesController.php
+++ b/src/Controller/ModulesController.php
@@ -463,10 +463,10 @@ class ModulesController extends AppController
     }
 
     /**
-     * Relation data load callig api `GET /:object_type/:id/related/:relation`
+     * Relation data load via API => `GET /:object_type/:id/related/:relation`
      *
-     * @param string|int $id the object identifier.
-     * @param string $relation the relating name.
+     * @param string|int $id The object ID.
+     * @param string $relation The relation name.
      * @return void
      */
     public function relatedJson($id, string $relation): void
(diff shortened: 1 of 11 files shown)


src/Controller/ModulesController.php
```diff
@@ -543,6 +557,33 @@ public function relationshipsJson($id, string $relation): void
...
+            return $defaultUrl;
+        }
+
+        $relationsSchema = $this->Schema->getRelationsSchema();
+        $types = $this->Modules->relatedTypes($relationsSchema, $relation);
+        if (count($types) === 1) {
+            return sprintf('/%s', $types[0]);
+        }
+
+        return '/objects?filter[type][]=' . implode('&filter[type][]=', $types);
```
didoda, line 584: if there are several types, url can be very long. I wonder whether BE api allow a shorter version of this kind of url, like:
`/objects?filter[type]=<comma separated types>`

didoda (approved):

src/Controller/ModulesController.php
stefanorosanelli: the alternative format you suggested is not supported right now by the API, but could be an improvement surely...

PR #448, opened 2020-09-16
Feat/history

DarioSacco:
Add `info` `restore` and `clone` features to history panel:
- `info` shows, in the side panel, detailed information about changes made to the object in that specific history entry
- `restore` restores the data from a history entry and places it in the form (without saving the object)
- `clone` does the same as `restore` but places the restored data in a new object (and in a new browser tab), calling the new endpoint `/clone/:id/history/:id`.

_Other changes_: now the object's `clone` button also opens the clone in a new browser tab.


src/Controller/ModulesController.php
```diff
@@ -430,6 +430,61 @@ public function clone($id): ?Response
...
+        $this->viewBuilder()->setTemplate('view');
+
+        $schema = $this->Schema->getSchema();
+        if (!is_array($schema)) {
+            $this->Flash->error(__('Cannot create abstract objects or objects without schema'));
+
+            return $this->redirect(['_name' => 'modules:list', 'object_type' => $this->objectType]);
+        }
+        try {
+            $historyResponse = $this->apiClient->get('/history', ['filter' => ['resource_id' => $id]]);
```
stefanorosanelli: Using this query string array you don't need `usort` and `array_slice` after...

```php
  'filter' => [
     'resource_id' => $id
     'resource_type' => 'objects',
     'id' => ['ge' => $historyId],
  ],
  'sort' => 'created',
]
```

stefanorosanelli: At least a `page_size` query should also be added, history could be long...

stefanorosanelli (commented):
* attribute values in "info" panel should be formatted somehow => dates are displayed in ISO-8601 format, select values should be _humanized_ etc etc (we could do it later)

* in `clone history` view many attributes are missing => `object` view array should also contain all schema properties (maybe with `null` values)

*  a minimal PHP test case should be added 

* couple of issues on `restore`
  - fields like `status` remain unchaged, `uname` becomes blank (if not set/changed by user should not change)
 - you can leave the page without usual alert `Changes you made may not be saved.`

src/Controller/ModulesController.php
DarioSacco: @stefanorosanelli for the `uname` point, is it fine to process it in an hard-coded way? It doesn't appear in the history creation entry

stefanorosanelli: @DarioSacco since `uname` has to be unique the safest thing to do is probably to unset it and force an automatic creation

PR #458, opened 2020-10-02
Issue/BSB-179-locations

nicolocarpignoli:
https://chialab.atlassian.net/browse/BSB-179

"Location" relation new view. This branch comes with modifies from the https://github.com/bedita/manager/pull/452 branch too.


src/Controller/ModulesController.php
```diff
@@ -354,6 +354,15 @@ protected function saveRelated(array $relatedData, string $id): void
...
             $relatedIds = (array)Hash::get($rel, 'relatedIds');
+
+            foreach ($relatedIds as $key => &$data) {
+                if (!empty($data['attributes'])) {
+                    $object = $this->apiClient->save($data['type'], array_merge([
+                        'id' => $data['id'] ?? null,
+                    ], $data['attributes']));
+                    $data['id'] = $object['data']['id'];
+                }
+            }
```
didoda: this saves all related objects. why? the save logic should be performed by following `$this->apiClient->{$method}($id, $this->objectType, $relation, $relatedIds)`

nicolocarpignoli: @edoardocavazza

edoardocavazza: @didoda Does the POST call to update relationships update entities if attributes are passed?

stefanorosanelli: It can be handy, actually: you can create or update related objects while updating a relation in one shot.
I would `unset($data['attributes'])` after save though.

src/Template/Element/Form/relation.twig
```diff
@@ -108,15 +108,13 @@
         {{ Form.control(relationName ~ 'removeRelated', {
             'type': 'hidden',
             'name': 'relations[' ~ relationName ~ '][removeRelated]',
-            'v-model': 'removedRelationsData'
```
stefanorosanelli: Why are we removing `'v-model': 'removedRelationsData'` and `'v-model': 'addedRelationsData'` ?
It affects object relations handling in general.

Also in `src/Template/Layout/js/app/components/relation-view/relation-view.js`

src/Controller/ModulesController.php
didoda: Fair enough. Maybe we could add proper test cases to `ModulesControllerTest::testSaveRelated`

src/Controller/ModulesController.php
```diff
@@ -350,6 +350,26 @@ protected function saveRelated(array $relatedData, string $id): void
...
+                if (!empty($data['attributes'])) {
+                    // avoid empty objects
+                    $saveObject = false;
+                    foreach ($data['attributes'] as $key => $field) {
+                        if ($key != 'status' && !empty($field)) {
+                            $saveObject = true;
+                        }
+                    }
+
+                    if ($saveObject == true) {
```
didoda: minor, suggested by scrutinizer

src/Controller/ModulesController.php
```diff
@@ -350,6 +350,26 @@ protected function saveRelated(array $relatedData, string $id): void
...
+                        $object = $this->apiClient->save($data['type'], array_merge([
+                            'id' => $data['id'] ?? null,
+                        ], $data['attributes']));
+                    }
+
+                    $data['id'] = $object['data']['id'];
```
didoda: minor, suggested by scrutinizer too.
It could be fixed adding `$object = [];` somewhere up in the code and using `$data['id'] = Hash::get($object, 'data.id');` in here.

src/Controller/ModulesController.php
```diff
@@ -350,6 +350,26 @@ protected function saveRelated(array $relatedData, string $id): void
             $method = (string)Hash::get($rel, 'method');
             $relation = (string)Hash::get($rel, 'relation');
             $relatedIds = (array)Hash::get($rel, 'relatedIds');
+
+            foreach ($relatedIds as $key => &$data) {
+                if (!empty($data['attributes'])) {
+                    // avoid empty objects
+                    $saveObject = false;
+                    foreach ($data['attributes'] as $key => $field) {
```
didoda: Using the same `$key` variable in cascade foreach:
```
foreach ($relatedIds as $key => &$data) { // line 354
[...]
    foreach ($data['attributes'] as $key => $field) { // line 358
```
Better using another variable name (`$k` or something else).

didoda (changes requested):

src/Controller/ModulesController.php
nicolocarpignoli: What exactly? 

You are saying to just use 
`if ($saveObject )` ?

src/Controller/ModulesController.php
nicolocarpignoli: fixed

src/Controller/ModulesController.php
nicolocarpignoli: why should we do that?

src/Controller/ModulesController.php
didoda:
```
It seems like you are loosely comparing two booleans. Considering using the strict comparison === instead.
```

`if ($saveObject === true)`

src/Controller/ModulesController.php
didoda: sorry, maybe you don't see `scrutinizer` suggestions:

```
The variable $object does not seem to be defined for all execution paths leading up to this point.
```

didoda (approved):

Issue #464, opened 2020-10-12
Relation panel - filter associable objects by type when relation.right is objects


didoda:
In relation panel, it's not possible to filter objects by object type when relation' schema right property is `objects`.

Actual bejaviour: filter by object type not available in "children" panel (for instance)

![children-panel](https://user-images.githubusercontent.com/2227145/95707516-0a0ac500-0c5a-11eb-80ca-90e636591c33.png)

Expected behaviour: filter by object type available in panel when schema right property is `objects`, like in relations with multiple object types involved

![relation-panel](https://user-images.githubusercontent.com/2227145/95707648-62da5d80-0c5a-11eb-9367-655d32a79811.png)

didoda:
done with #465

PR #465, opened 2020-10-12
feat: concrete types in relation view and panel

didoda:
This provides #464


src/Controller/Component/SchemaComponent.php
```diff
@@ -101,6 +101,30 @@ function () use ($type) {
...
+     * @param array $schema The schema
+     * @param array $relationships The relationships
+     * @param array $types The types to use instead of 'objects'
+     * @return array
+     */
+    public function relationsSchema(array $schema, array $relationships, array $types): array
+    {
+        $relationsSchema = array_intersect_key($schema, $relationships);
+        foreach ($relationsSchema as &$relSchema) {
+            if (!array_key_exists('right', $relSchema)) {
```
stefanorosanelli: or something like:
```php
if (in_array('objects', (array)Hash::get($reslSchema, 'right')) {
   $relSchema['right'] = $types;
}
```

src/Controller/Component/ModulesComponent.php
```diff
@@ -459,7 +459,9 @@ public function prepareQuery(array $query): array
     public function setupRelationsMeta(array $schema, array $relationships, array $order = []): void
     {
         // relations between objects
-        $relationsSchema = array_intersect_key($schema, $relationships);
+        $types = $this->objectTypes(false);
+        sort($types);
+        $relationsSchema = $this->getController()->Schema->relationsSchema($schema, $relationships, $types);
```
stefanorosanelli: Since `relationsSchema` is only used in `ModulesComponent` for now we could add it to this class as `protected`.

We have other `abstract` classes like `media` and possibly others where we need to extract `concrete` types. 
We should have this information already in cached relations schema (next step), then `relationsSchema()` will not be necessary anymore.

didoda: so, do you suggest to move `relationSchema` method from `SchemaComponent` to `ModulesComponent` (+move testRelationSchema)?

stefanorosanelli: yes, for now

in a future PR we may delete it probably (see above)

tests/TestCase/Controller/Component/ModulesComponentTest.php
```diff
@@ -1107,6 +1110,8 @@ public function setupRelationsProvider()
      */
     public function testSetupRelationsMeta(array $expected, array $schema, array $relationships, array $order = [])
     {
+        $registry = $this->Modules->getController()->components();
+        $this->Modules->getController()->Schema = $registry->load(SchemaComponent::class);
```
stefanorosanelli: Is this still necessary?

didoda: not necessary. removing

stefanorosanelli (approved):

PR #467, opened 2020-10-14
coords geojson form field

didoda:
This provides a custom handling for coords property.

Bonus: move `\App\Core\Utility\Form` logic in separate classes and refactor.


src/Core/Utility/Form/Control.php
```diff
@@ -0,0 +1,236 @@
...
+     * @param string $type Control type.
+     * @param mixed|null $value Property value.
+     * @return array
+     */
+    public static function control(array $schema, string $type, $value): array
+    {
+        if (!in_array($type, Control::CONTROL_TYPES)) {
+            return compact('type', 'value');
+        }
+        $method = Form::getMethod(Control::class, sprintf('%s', $type));
```
le0m: no need for `sprintf`

src/Core/Utility/Form/Options.php
```diff
@@ -0,0 +1,224 @@
...
+     * @param string $name The field name.
+     * @param mixed|null $value The field value.
+     * @return array
+     */
+    public static function customControl(string $name, $value): array
+    {
+        if (!in_array($name, Options::CUSTOM_CONTROLS)) {
+            return [];
+        }
+        $method = Form::getMethod(Options::class, sprintf('%s', $name));
```
le0m: no need for `sprintf`

didoda: good point ^_^

src/Core/Utility/Form/Control.php
```diff
@@ -0,0 +1,236 @@
...
+ * See LICENSE.LGPL or <http://gnu.org/licenses/lgpl-3.0.html> for more details.
+ */
+
+namespace App\Core\Utility\Form;
+
+use Cake\Core\Configure;
+use Cake\Utility\Hash;
+use Cake\Utility\Inflector;
+
+class Control
```
stefanorosanelli: Good job! A class refactor was needed and class decomposition is fine.

* Con we add a comment line on each class to describe its purpose?
* why `\App\Core\Utility\Form` namespace and not `\App\Utility\Form` or `\App\Form` ?

didoda: > Con we add a comment line on each class to describe its purpose?

class comments coming soon ;)

> why \App\Core\Utility\Form namespace and not \App\Utility\Form or \App\Form ?

good point. namespace `\App\Form` could be fine. I'll apply it soon

stefanorosanelli (changes requested):
Seems ok, good job!

Just one issue: if `coords` property is set in `core` group the input text is not displayed 

To reproduce use this conf in `Properties`

```php
        'locations' => [
            'view' => [
                'core' => [
                    'title',
                    'coords',
                ],
            ],
        ],
```
Screenshot_20201015_105032

didoda:
> Seems ok, good job!
> 
> Just one issue: if `coords` property is set in `core` group the input text is not displayed

ok. it's probably a webpack related problem.
`CoordinatesView` component is enabled in `property-view` (the view of the non-core blocks).
We should enable it in core too. I'll take a look on this

stefanorosanelli (approved):

PR #478, opened 2020-11-24
Property Helper and control handlers for plugins

didoda:
This provides a refactor for the purpose of enabling custom controls defined in plugins or elsewhere (not in manager core).

You may have a configuration like this in `config/app.php` : 

```php
    'Control' => [
        'handlers' => [
            'my_object_type' => [
                'my_attribute' => [
                    'class' => 'MyPlugin\Path\To\SomeClass',
                    'method' => 'someMethod',
                ],
            ],
        ],
    ],
``` 

Or in a plugin's bootstrap you may define: 

```php
use Cake\Core\Configure;

Configure::write(
    'Control.handlers',
    array_merge(
        (array)\Cake\Core\Configure::read('Control.handlers'),
        [
             'my_object_type' => [
                'my_attribute' => [
                    'class' => 'MyPlugin\Path\To\SomeClass',
                    'method' => 'someMethod',
                ],
            ],
        ]
    )
);
```
We expect that field `my_object_type.my_attribute` is rendered through `MyPlugin\Path\To\SomeClass::someMethod` call.

This also add a `PropertyHelper` to allow a single row statement to build a property control for an object attribute.

Bonus: move some tests from `Core/Utility/Form` to `Form` namespace. Minor fixes on tests, proper order of assert parameters.


src/View/Helper/PropertyHelper.php
```diff
@@ -0,0 +1,60 @@
...
+    /**
+     * View the property
+     *
+     * @param string $key The property key
+     * @param mixed|null $value The property value
+     * @param array $schema The property schema
+     * @param array $options The form element options, if any
+     * @return string
+     */
+    public function view(string $key, $value, array $schema, array $options = []): string
```
stefanorosanelli: We could remove `$schema` from arguments list and use `schema` view value using `$this->_View->get('schema')`

src/Form/Control.php
```diff
@@ -38,12 +38,21 @@ class Control
      */
     public static function control(array $schema, string $type, $value): array
     {
-        if (!in_array($type, Control::CONTROL_TYPES)) {
+        $handlers = Configure::read('Control.handlers', []);
```
stefanorosanelli: Why not have a configuration indexed by property type?
This way we could have a simpler code:

```php
$handler = Configure::read('Control.handlers.' . $type);
if (!empty($handler)) {
        return call_user_func_array([$handler['class'], $handler['method']], [$value, $schema]);
}
```
remaining code method is unchanged

src/View/Helper/PropertyHelper.php
```diff
@@ -0,0 +1,78 @@
...
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * See LICENSE.LGPL or <http://gnu.org/licenses/lgpl-3.0.html> for more details.
+ */
+namespace App\View\Helper;
+
+use Cake\Core\Configure;
+use Cake\Core\Plugin;
```
stefanorosanelli: some unused imports...

stefanorosanelli (approved):

PR #489, opened 2020-12-18
Unlock fields, by '<moduleName>.<fieldName>` setting

didoda:
In some contexts (i.e. plugins customizations of fields rendering) we need to unlock specific fields.
Usually we use `{{ Form.unlockField(fieldName) }}` in `twig` templates.
When no twig templates are involved, submit causes a Security error.

This provides a workaround.
Set a config to unlock fields as follows:
```
Configure::write(
    'animals.unlockedFields',
    array_merge(
        (array)Configure::read('animals.unlockedFields'),
        [
            'eyes', 'legs',
        ]
    )
);
```
`ModulesController::beforeRender` will set `unlockedField` `['eyes', 'legs']`, when current module is `animals`.

### DIFF
--- a/src/Controller/ModulesController.php
+++ b/src/Controller/ModulesController.php
@@ -14,6 +14,7 @@ namespace App\Controller;
 
 use App\Core\Exception\UploadException;
 use BEdita\SDK\BEditaClientException;
+use Cake\Core\Configure;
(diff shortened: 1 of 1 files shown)


didoda:
not necessary. better using custom elements to provide field's unlock using `twig` templates.

PR #494, opened 2021-01-04
Phpspreadsheet

didoda:
This provides export to csv|ods|xlsx using Phpspreadsheet (https://phpspreadsheet.readthedocs.io/en/latest/).

Available formats can be set in configuration as follows:
```
'Export' => [
    'limit' => 10000,
    'formats' => [
        'CSV' => 'csv',
        'Open Document' => 'ods',
        'MS Excel' => 'xlsx'
    ],
    'default' => 'xlsx'
],
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,6 @@
         "cakephp/plugin-installer": "^1.1",
         "josegonzalez/dotenv": "2.*",
         "mobiledetect/mobiledetectlib": "2.*",
-        "mpdf/mpdf": "^8.0",
         "phpoffice/phpspreadsheet": "^1.15",
         "wikimedia/composer-merge-plugin": "^1.4"
     },
--- a/composer.json
+++ b/composer.json
@@ -22,11 +22,13 @@
     ],
     "require": {
         "php": ">=7.1",
-        "bedita/web-tools": "^1.5.0",
         "bedita/i18n": "^1.5",
+        "bedita/web-tools": "^1.5.0",
         "cakephp/plugin-installer": "^1.1",
         "josegonzalez/dotenv": "2.*",
         "mobiledetect/mobiledetectlib": "2.*",
+        "mpdf/mpdf": "^8.0",
+        "phpoffice/phpspreadsheet": "^1.15",
         "wikimedia/composer-merge-plugin": "^1.4"
     },
     "require-dev": {
--- a/src/Controller/Component/ExportComponent.php
+++ b/src/Controller/Component/ExportComponent.php
@@ -1,0 +1,186 @@
+<?php
+namespace App\Controller\Component;
+
+use Cake\Controller\Component;
+use Cake\Controller\ComponentRegistry;
+use Cake\Http\Response;
+use Cake\Utility\Hash;
+use PhpOffice\PhpSpreadsheet\IOFactory;
+use PhpOffice\PhpSpreadsheet\Spreadsheet;
+use PhpOffice\PhpSpreadsheet\Worksheet\PageSetup;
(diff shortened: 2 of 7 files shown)


composer.json
```diff
@@ -22,11 +22,13 @@
     ],
     "require": {
         "php": ">=7.1",
-        "bedita/web-tools": "^1.5.0",
         "bedita/i18n": "^1.5",
+        "bedita/web-tools": "^1.5.0",
         "cakephp/plugin-installer": "^1.1",
         "josegonzalez/dotenv": "2.*",
         "mobiledetect/mobiledetectlib": "2.*",
+        "mpdf/mpdf": "^8.0",
```
stefanorosanelli: is `mpdf` needed without `pdf` export?

src/Controller/Component/ExportComponent.php
```diff
@@ -0,0 +1,166 @@
...
+namespace App\Controller\Component;
+
+use Cake\Controller\Component;
+use Cake\Controller\ComponentRegistry;
+use Cake\Http\Response;
+use Cake\Utility\Hash;
+use PhpOffice\PhpSpreadsheet\IOFactory;
+use PhpOffice\PhpSpreadsheet\Spreadsheet;
+use PhpOffice\PhpSpreadsheet\Worksheet\PageSetup;
+use PhpOffice\PhpSpreadsheet\Writer\Pdf\Mpdf;
```
stefanorosanelli: `Mpdf` unused

src/Controller/ExportController.php
```diff
@@ -29,24 +31,48 @@ class ExportController extends AppController
...
      *
-     * @return \Cake\Http\Response
+     * @return \Cake\Http\Response|null
      */
-    public function export(): Response
+    public function export(): ?Response
     {
+        // allow only specific export formats: csv, ods, xlsx
+        $format = (string)$this->request->getData('format');
+        if (!in_array($format, ['csv', 'ods', 'xlsx'])) {
```
stefanorosanelli: use a `const ALLOWED_FORMATS` instead of `['csv', 'ods', 'xlsx']` array? 

format list could change, also using plugins for instance

src/Controller/ExportController.php
```diff
@@ -29,24 +31,48 @@ class ExportController extends AppController
...
 
-        // load csv data for objects by object type and ids
-        $rows = $this->csvRows($data['objectType'], $ids);
+        // load data for objects by object type and ids
+        $rows = $this->rows($data['objectType'], $ids);
+
+        // create spreadsheet and return as download
+        $spreadsheet = $this->Export->spreadsheet($rows);
+        $filename = sprintf('%s_%s.%s', $data['objectType'], date('Ymd-His'), $format);
+        $data = $this->Export->{$format}($spreadsheet, $filename);
```
stefanorosanelli: I'd prefer an `ExportComponent::format($format, $rows, $filename)` and no `public` `spreadsheet()` method

 * in `ExportComponent::format()` we could load other formats via plugin not requiring `spreadsheet()` method
 * we could simply add a `json` format easily 
 
Also an `ExportComponent::checkFormat()` could be used, only `ExportComponent` knows the available formats?
Or an `ExportComponent::ALLOWED_FORMATS` array for now.

composer.json
didoda: no. we can remove the dependency

src/Controller/ExportController.php
didoda: ok. some refactor soon

stefanorosanelli (approved):

PR #506, opened 2021-01-19
Fix/filter status

didoda:
This fixes "status filter" bug in modules index.

= actual behavior =

When status filter is "on|off|draft", changing it to "All" doesn't work: "on|off|draft" status filter can't be removed.

= expected behavior =

When status filter is "All", filter[status] is removed from url, and no status filter is applied.

### DIFF
--- a/src/Controller/AppController.php
+++ b/src/Controller/AppController.php
@@ -382,7 +382,7 @@ class AppController extends Controller
             $session->write($sessionKey, $this->request->getQueryParams());
 
             return null;
-        } else { // empty? remove sessione key
+        } else { // empty? remove session key
             $session->delete($sessionKey);
         }
 
--- a/src/Controller/AppController.php
+++ b/src/Controller/AppController.php
@@ -377,19 +377,11 @@ class AppController extends Controller
             return $this->redirect((string)$this->request->getUri()->withQuery(''));
         }
 
-        // _search === 1 => request from button Search
-        if ($this->request->getQuery('_search') === '1') {
-            $session->delete($sessionKey);
-            $q = $this->request->getQueryParams();
-            unset($q['_search']);
-            $query = http_build_query($q, '', '&', PHP_QUERY_RFC3986);
-
-            return $this->redirect((string)$this->request->getUri()->withQuery($query));
-        }
-
         // write request query parameters (if any) in session
-        if (!empty($this->request->getQueryParams())) {
-            $session->write($sessionKey, $this->request->getQueryParams());
+        $params = $this->request->getQueryParams();
+        if (!empty($params)) {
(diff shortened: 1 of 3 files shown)


src/Controller/AppController.php
```diff
@@ -382,6 +382,8 @@ protected function applySessionFilter(): ?Response
             $session->write($sessionKey, $this->request->getQueryParams());
 
             return null;
+        } else { // empty? remove session key
+            $session->delete($sessionKey);
```
stefanorosanelli: This way session filter is not working... every URL with empty query string will cause session key removal

src/Controller/AppController.php
```diff
@@ -377,6 +377,16 @@ protected function applySessionFilter(): ?Response
             return $this->redirect((string)$this->request->getUri()->withQuery(''));
         }
 
+        // _search === 1 => request from button Search
+        if ($this->request->getQuery('_search') === '1') {
+            $session->delete($sessionKey);
+            $q = $this->request->getQueryParams();
+            unset($q['_search']);
+            $query = http_build_query($q, null, '&', PHP_QUERY_RFC3986);
```
stefanorosanelli: should be `http_build_query($q, '', '&', PHP_QUERY_RFC3986);`

src/Controller/AppController.php
```diff
@@ -377,6 +377,16 @@ protected function applySessionFilter(): ?Response
             return $this->redirect((string)$this->request->getUri()->withQuery(''));
         }
 
+        // _search === 1 => request from button Search
```
stefanorosanelli: I propose:

```
        $params = $this->request->getQueryParams();
        if (!empty($params)) {
            unset($params['_search']);
            $session->write($sessionKey, $params);

            return null;
        }
```
Replace from line 380 to 394

src/Controller/AppController.php
```diff
@@ -378,15 +378,17 @@ protected function applySessionFilter(): ?Response
...
+            unset($params['_search']);
+            $session->write($sessionKey, $params);
 
             return null;
         }
 
         // read request query parameters from session and redirect to proper page
         if ($session->check($sessionKey)) {
-            $query = http_build_query($session->read($sessionKey), null, '&', PHP_QUERY_RFC3986);
+            $query = http_build_query($session->read($sessionKey), '', '&', PHP_QUERY_RFC3986);
```
stefanorosanelli: To avoid infinite loops I suggest:

```php
        $params = (array)$session->read($sessionKey);
        if (!empty($params)) {
            $query = http_build_query($params, '', '&', PHP_QUERY_RFC3986);

            return $this->redirect(....);
        }
```

PR #509, opened 2021-01-22
Expired token exception

didoda:
This forces logout redirect when dashboard index "recent items" throws a 401 Unauthorized exception.
Typical scenario is "expired jwt".


src/Controller/DashboardController.php
```diff
@@ -38,7 +38,14 @@ public function initialize(): void
     public function index(): void
     {
         $this->request->allowMethod(['get']);
-        $this->set('recentItems', $this->recentItems());
+        try {
```
stefanorosanelli, line 41: Session expiration cases is handled in `AppController::beforeFilter()` - user is redirected lo login page with post-login redirect url.
This is not yet working where in-page calls are performed (like right panel for relations in object view).

JWT token expiration is handled by PHP SDK directly with renew.

If there's a problem or bug in token or session expiration it should be solved there (`AppController` or PHP SDK)

didoda:
If API handles properly token expired, issue is not reproducible. This PR is not necessary.

PR #510, opened 2021-01-27
filter box handle search change and submit

### LINKED ISSUE #401: text searches in index filters less than 3 characters long
In a generic module index, text searches that are less than 3 characters long throw an exception (from the API).
After the error, the user is out of the module, back to the Dashboard, and all also the current filters are lost.

1. at least, the error should be managed in the module;
2. probably the form submit should not be enabled if the search string is > 0 and <3 characters long;

p.s.: for this error, is `400 not found` a good answer from the API?

didoda:
This fixes #401 showing a warning dialog when search text is too short.
Bonus: some translations using ttag in filter-box.

### DIFF
--- a/src/Template/Layout/js/app/components/filter-box.js
+++ b/src/Template/Layout/js/app/components/filter-box.js
@@ -17,6 +17,7 @@
 
 import { DEFAULT_PAGINATION, DEFAULT_FILTER } from 'app/mixins/paginated-content';
 import merge from 'deepmerge';
+import { t } from 'ttag';
 
 export default {
     components: {
@@ -26,11 +27,11 @@ export default {
     props: {
         objectsLabel: {
             type: String,
-            default: "Objects"
+            default: t`Objects`
         },
         placeholder: {
             type: String,
-            default: "Search"
+            default: t`Search`
         },
         showFilterButtons: {
             type: Boolean,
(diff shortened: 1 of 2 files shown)


src/Template/Layout/js/app/components/filter-box.js
```diff
@@ -161,17 +162,66 @@ export default {
...
+            dialog.cancelMessage = '';
+            const confirmCallback = (newVal) => {
+                search.value = newVal.trim();
+                if (!this.queryFilter) {
+                    this.queryFilter = {};
+                }
+                this.queryFilter.q = search.value;
+                dialog.hide();
+            };
+            dialog.prompt(t`Search text too short. Minimum length is 3. Retry`, search.value, confirmCallback);
```
stefanorosanelli: [minor] search value in dialog could be read-only or simply added to the message - otherwise you may think of change it and relaunch search from the dialog (value is editable...)

stefanorosanelli (approved):

src/Template/Layout/js/app/components/filter-box.js
didoda: you can change it with prompt, actually

didoda: let's simplify. warning instead of prompt

PR #511, opened 2021-01-28
Status and modified columns when available, in index

didoda:
In modules index "status" and "modified" column are shown only when data contain them.

Scenario: custom module could load a list of objects without "attributes.status" and|or "meta.modified": with this fix, that module index would hide columns without relative data.

### DIFF
--- a/src/Template/Pages/Modules/index.twig
+++ b/src/Template/Pages/Modules/index.twig
@@ -20,7 +20,7 @@
 
             <div class="list-objects">
                 {% if objects %}
-                    {% element 'Modules/index_table_header' { 'object': objects[0] } %}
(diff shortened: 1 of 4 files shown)


src/Template/Pages/Modules/index.twig
```diff
@@ -20,7 +20,7 @@
 
             <div class="list-objects">
                 {% if objects %}
-                    {% element 'Modules/index_table_header' %}
+                    {% element 'Modules/index_table_header' { 'object': objects[0] } %}
```
stefanorosanelli: [super minor] maybe `firstObject` or `refObject` instead of `object` here and in `Modules/index_table_header.twig` to avoid confusion with `object/objects` vars else where

stefanorosanelli (approved):

PR #527, opened 2021-03-15
Save related data for children, removal case

### LINKED ISSUE #525: object view - save children, do not remove items
In an object edit page, in the box "children" we can add or remove items.
"Removed items" appear with a gray background.
On save, system doesn't remove the "Removed items" from object's children.
Expected behaviour: remove "Removed items" from object's children.

didoda:
This fixes #525


stefanorosanelli (approved):

nicolocarpignoli:
Works with objects but gives "403 - Not Found" error when removing folder as children of another folder.

PR #533, opened 2021-03-22
Create a dialog service component

edoardocavazza:
This PR introduce a constructable Dialog component and some helper methods to replace the singleton dialog element.

### DIFF
--- a/src/Template/Layout/js/app/components/dialog/dialog.js
+++ b/src/Template/Layout/js/app/components/dialog/dialog.js
@@ -58,7 +58,7 @@ export const Dialog = Vue.extend({
             this.dialogType = '';
             this.isOpen = false;
 
-            if (destroy ?? this.destroyOnHide) {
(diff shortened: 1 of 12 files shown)


src/Template/Layout/js/app/components/dialog/dialog.js
```diff
@@ -43,40 +43,83 @@ export const BeditaDialog = {
...
+            if (this.$el.parentNode !== root) {
+                root.appendChild(this.$el);
+            }
         },
-        hide() {
+        hide(destroy = false) {
             this.dialogType = '';
             this.isOpen = false;
+
+            if (this.destroyOnHide || destroy) {
```
DarioSacco: Isn't it better to invert this check? If you pass `false` from outside, checking `this.destroyOnHide` (that defaults to `true`), the dialog will be still destroyed

didoda (approved):

PR #538, opened 2021-04-01
style: show uname on list, list view as default for "uploadable" relations

nicolocarpignoli:
Changelog:

* show list view as default 
* show uname prop. on list view

### DIFF
--- a/src/Template/Element/Form/relation.twig
+++ b/src/Template/Element/Form/relation.twig
@@ -1,6 +1,7 @@
 <relation-view inline-template
     :relation-data='{{ relationSchema|json_encode|raw }}'
     :list-view=listView
+    accepted-length={{ acceptedLength }}
     relation-name={{ relationName }}
     :pre-count={{ preCount }}
     config-paginate-sizes={{ config('Pagination.sizeAvailable')|json_encode()|raw }}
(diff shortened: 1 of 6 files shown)


src/Template/Layout/js/app/components/relation-view/relation-view.js
```diff
@@ -86,6 +82,9 @@ export default {
     },
 
     computed: {
+        listView() {
+            return this.relationName != 'poster' && this.relationName != 'attach';
```
stefanorosanelli: instead of using relation names that are dynamic we could define this behavior looking for `uploadable` media types in related types 

See: https://github.com/bedita/manager/blob/master/src/Template/Element/Form/relation.twig#L78 

Here we decide to show a drop area, but the logic is the same. We could probably define a Twig boolean variable and pass it as a Vue component `prop`

src/Template/Element/Form/relation.twig
```diff
@@ -75,7 +76,6 @@
         {% if Perms.canSave() %}
         <div class="mt-5">
             {# DROP FILES #}
-            {% set acceptedTypes = uploadable | filter((item) => item in relationSchema.right) | join(', ') %}
             {% if acceptedTypes|length %}
```
stefanorosanelli: should be  ```{% if acceptedLength %}``` now ?

drop area is not showing where expected as a consequence

stefanorosanelli (changes requested):
`card` mode seems selected for related objects near count number, whereas `list` mode is used as default

Screenshot_20210406_101445

stefanorosanelli (approved):

src/Template/Element/Form/relations.twig
```diff
@@ -2,10 +2,12 @@
 <section class="relations">
 
     {% for relationName, relationLabel in relations %}
+        {% set acceptedTypes = uploadable | filter((item) => item in relationsSchema[relationName].right) | join(', ') %}
```
stefanorosanelli: in this context we could probably use something like `uploadableTypes` and `uploadableNum` (number of uploadable types in related types)

PR #543, opened 2021-04-12
BEM-11 default columns for media objects index page

didoda:
This provides default columns for media objects index page:

 - title
 - file_name
 - mime_type
 - file_size (friendly format, i.e.: `:num` Kb, Mb, Gb, Tb)

This applies to index page for:

 - audio
 - files
 - images
 - media
 - videos

This requires https://github.com/bedita/web-tools/pull/38 / https://github.com/bedita/web-tools/pull/39


src/View/Helper/PropertyHelper.php
```diff
@@ -73,4 +73,28 @@ public function schema(string $name): array
...
     }
+
+    /**
+     * Get object property value by object and property.
+     *
+     * @param array $object The object
+     * @param string $property The property
+     * @return string
+     */
+    public function objectValue(array $object, string $property): string
```
stefanorosanelli: [minor] rename to `value()` - could be used by resources and objects 
also `array $resource` or `array $item` as argument

src/Controller/Component/PropertiesComponent.php
```diff
@@ -152,7 +160,11 @@ public function viewGroups(array $object, string $type): array
      */
     public function indexList(string $type): array
```
stefanorosanelli: I'd prefer to leave `indexList()` unchanged and create default `index` lists in `config/properties` where `DefaultProperties` are defined.

Example:

```
        'images' => [
            'index' => [
                'title',
                'file_name',
                'mime_type',
                'file_size',
            ],
        ],
```

src/Controller/ModulesController.php
```diff
@@ -96,7 +96,7 @@ public function index(): ?Response
 
         $this->ProjectConfiguration->read();
 
-        $objects = (array)$response['data'];
+        $objects = $this->Modules->objects((array)$response);
```
stefanorosanelli: We could introduce an `ApiFormatter` component (via WebTools) to embed included resources/objects in `relationships` section of the response and avoid this method. 
@didoda more details via call later

src/Template/Element/Modules/index_table_row.twig
```diff
@@ -26,7 +26,7 @@
                 </div>
             {% else %}
                 <div class="{{ prop }}-cell" untitled-label="{{ __('Untitled') }}">
-                    {% set propVal = object.attributes[prop] ?? object.meta[prop] %}
+                    {% set propVal = Property.objectValue(object, prop) %}
```
stefanorosanelli: we could probably use `Schema.format` inside `Property` method call & introduce a `bytes` primitive type to format `file_size` - more details later also on this subject

src/View/Helper/PropertyHelper.php
didoda: ok. renaming done ;)

src/Controller/Component/PropertiesComponent.php
didoda: Good point. Used `DefaultProperties` instead.

src/Controller/ModulesController.php
didoda: in the direction of your proposal: https://github.com/bedita/web-tools/pull/38

when that PR will be merged, we could update this PR as necessary to use the `ApiFormatter`.

stefanorosanelli (approved):

PR #546, opened 2021-04-15
fix: show error when format of exporter is not allowed

paolosantarsiero:
Bug: When try to export a file and the format is not allowed it return error 500.
Fix: Show a message that format is not allowed.


src/Controller/ExportController.php
```diff
@@ -56,14 +56,19 @@ public function initialize(): void
...
-        }
+
         // check request (allowed methods and required parameters)
         $data = $this->checkRequest([
             'allowedMethods' => ['post'],
             'requiredParameters' => ['objectType'],
         ]);
+
+        if (!$this->Export->checkFormat($format)) {
+            $this->Flash->error('Il formato selezionato non è disponibile!');
```
stefanorosanelli: We use english for labels and messages with `__()` function to enable translations via gettext
It should be:

```php
$this->Flash->error(_('Selected format....'));
```

src/Controller/ExportController.php
```diff
@@ -56,14 +56,19 @@ public function initialize(): void
...
+
+            return $this->redirect(['_name' => 'modules:list', 'object_type' => $data['objectType']]);
```
stefanorosanelli: Is there a test case for these lines?

stefanorosanelli (changes requested):

composer.json
```diff
@@ -28,7 +28,7 @@
         "josegonzalez/dotenv": "2.*",
         "mobiledetect/mobiledetectlib": "2.*",
         "phpoffice/phpspreadsheet": "^1.15",
-        "wikimedia/composer-merge-plugin": "^1.4"
+        "wikimedia/composer-merge-plugin": "^2.0"
```
didoda: This change is not related to the fix (it's related to the use of composer2)

src/Controller/ExportController.php
```diff
@@ -56,14 +56,19 @@ public function initialize(): void
...
         // check request (allowed methods and required parameters)
         $data = $this->checkRequest([
             'allowedMethods' => ['post'],
             'requiredParameters' => ['objectType'],
         ]);
+
+        if (!$this->Export->checkFormat($format)) {
+            $this->Flash->error(__('Format not allowed'));
+
+            return $this->redirect(['_name' => 'modules:list', 'object_type' => $data['objectType']]);
```
stefanorosanelli: safer redirect is probably this (we can't be sure where the request originated)

```php
return $this->redirect($this->referer());
```

composer.json
stefanorosanelli: right, but why do we have this diff since in `master` we have already `"wikimedia/composer-merge-plugin": "^2.0.1"` ? :thinking:

didoda: some merge conflict. resolved with a push...

didoda (approved):

tests/TestCase/Controller/ExportControllerTest.php
```diff
@@ -156,6 +156,28 @@ public function testExport(): void
...
+        $this->Export = new ExportController(
+            new ServerRequest([
+                'environment' => ['REQUEST_METHOD' => 'POST'],
+                'params' => ['objectType' => 'proms'],
+                'post' => ['ids' => '655', 'objectType' => 'proms', 'format' => ''],
+            ])
+        );
+
+        // call export.
+        $response = $this->Export->export();
```
stefanorosanelli, line 176: [minor]
```
static::assertEquals(302, $response->getStatusCode());
```

src/Locale/en_US/default.po
```diff
@@ -191,6 +191,9 @@ msgstr ""
 msgid "Export Filtered"
 msgstr ""
 
+msgid "Format not allowed"
+msgstr "Format choosen is not available"
```
stefanorosanelli: We usually don't _translate_ messages in this locale

use `"Format choosen is not available"` directly as message string instead of `"Format not allowed"` and update references

stefanorosanelli (approved):

PR #549, opened 2021-04-26
FIX: save changed order of children folders

le0m:
This PR fixes a problem when changing order of children folders from within a folder in the Manager UI. The new order was not saved because the parameter `meta` of the relation wasn't sent to the APIs.

### DIFF
--- a/src/Controller/Component/ModulesComponent.php
+++ b/src/Controller/Component/ModulesComponent.php
@@ -670,6 +670,9 @@ class ModulesComponent extends Component
             }
             // invert relation call => use 'parent' relation on children folder
             $data = compact('id') + ['type' => 'folders'];
+            if (Hash::check($item, 'meta')) {
+                $data += ['meta' => Hash::check($item, 'meta')];
+            }
             $apiClient->replaceRelated($relId, 'folders', 'parent', $data);
         }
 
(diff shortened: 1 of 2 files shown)


src/Controller/Component/ModulesComponent.php
```diff
@@ -670,6 +670,9 @@ protected function folderChildrenRelated(string $id, array $relatedIds)
             }
             // invert relation call => use 'parent' relation on children folder
             $data = compact('id') + ['type' => 'folders'];
+            if (Hash::check($item, 'meta')) {
+                $data += ['meta' => Hash::get($item, 'meta')];
```
didoda, line 674: to make scrutinizer happy
```suggestion
                $data += ['meta' => (array)Hash::get($item, 'meta')];
```

didoda (approved):

src/Controller/Component/ModulesComponent.php
```diff
@@ -670,6 +670,9 @@ protected function folderChildrenRelated(string $id, array $relatedIds)
             }
             // invert relation call => use 'parent' relation on children folder
             $data = compact('id') + ['type' => 'folders'];
+            if (Hash::check($item, 'meta')) {
```
didoda:
```suggestion
            if (Hash::check((array)$item, 'meta')) {
```
to make scrutinizer happy

PR #565, opened 2021-05-24
feat: refactor modules

didoda:
This moves some logic from `ModulesController` to new components `Query` and `Thumbs`

Bonus: more tests for `ModulesController`.


src/Controller/Component/ThumbsComponent.php
```diff
@@ -0,0 +1,86 @@
...
+            return;
+        }
+
+        // extract ids of objects
+        $ids = (array)Hash::extract($response, 'data.{n}[type=/images|videos/].id');
+        if (empty($ids)) {
+            return;
+        }
+
+        $getThumbs = function (array $ids): ?array {
```
stefanorosanelli: method seems very long - why not create a `protected function getThumbs(array $ids)` ?

src/Controller/Component/ThumbsComponent.php
```diff
@@ -0,0 +1,86 @@
...
+            try {
+                $params = $this->getController()->getRequest()->getQueryParams();
+                $query = $this->getController()->Query->prepare($params);
+                $url = sprintf('/media/thumbs?%s', http_build_query([
```
stefanorosanelli: Could we use the new `ThumbHelper` cache?

src/Controller/ModulesController.php
```diff
@@ -205,11 +189,10 @@ public function uname($id): ?Response
         try {
             $response = $this->apiClient->get(sprintf('/objects/%s', $id));
         } catch (BEditaClientException $e) {
-            if ($e->getCode() === 404) {
-                $error = sprintf(__('Resource "%s" not found', true), $id);
-            } else {
-                $error = sprintf(__('Resource "%s" not available. Error: %s', true), $id, $e->getMessage());
-            }
+            $msg = $e->getMessage();
```
stefanorosanelli, line 192: [super-minor] we usually prefer `if/else` over ternary operator because it's easier to debug ... not a big deal here...

src/Controller/ModulesController.php
```diff
@@ -410,7 +393,7 @@ public function delete(): ?Response
                     return $this->redirect(['_name' => 'modules:view', 'object_type' => $this->objectType, 'id' => $this->request->getData('id')]);
                 }
 
-                return $this->redirect(['_name' => 'modules:view', 'object_type' => $this->objectType]);
+                return $this->redirect(['_name' => 'modules:view', 'object_type' => $this->objectType, 'id' => $id]);
```
stefanorosanelli, line 396: is this a fix? `'id'` was missing here?

src/Controller/Component/ThumbsComponent.php
didoda: can try it

src/Controller/ModulesController.php
didoda: if/else create 2 paths, bad for coverage, good for readability. I choosed better coverage over readability

src/Controller/ModulesController.php
didoda: it's a fix. the other path would be unreachable

src/Controller/Component/ThumbsComponent.php
didoda: Manager uses `"bedita/web-tools": "^1.6.0"`, where ThumbHelper cache is not available

stefanorosanelli (approved):

PR #568, opened 2021-05-26
Refactor/new sidebar

qwerg:
PR with various changes to get closer to the original UI UX project
BE4-cms-MR

### DIFF
--- a/src/Template/Pages/Dashboard/index.twig
+++ b/src/Template/Pages/Dashboard/index.twig
@@ -21,9 +21,18 @@
                     <span>{{ title }}</span>
                 </a>
             {% endfor %}
-            </div>
 
-            <div class="dashboard-items dashboard-utility">
+            </div>
+        </section>
+
+        <section class="dashboard-section">
+        
+            {# <header>
+                <h2>{{ __('Preferences and Tools') }}</h2>
+            </header> #}
+
+            <div class="dashboard-items">
+
                 {% if modules.publications %}
                     {{ Html.link(__('Publications'),
                         { '_name': 'modules:list', 'object_type': 'publications' },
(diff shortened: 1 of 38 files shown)


src/Template/Pages/Dashboard/index.twig
```diff
@@ -40,25 +49,21 @@
                     { '_name': 'modules:list', 'object_type': 'objects' },
                     { 'title': __('All objects'), 'class': 'dashboard-item has-background-module-objects icon-th-large-1', } )|raw }}
 
-                {# <a href="#" class="dashboard-item has-background-black icon-list-1" disabled>{{ __('All categories') }}</a> #}
+                <a href="#" class="dashboard-item has-background-module-categories icon-list-1">{{ __('All categories') }}</a>
```
stefanorosanelli, line 53: `All categories` and `All tags` are not available yet - better keep both anchors commented

stefanorosanelli: Note: we have `Categories` under `Data modeling` and it's ok to handle them there.

If we want to have a standalone `All categories` module to be used by non-admin users we will probably clone it or (better) prepare two possible views.

stefanorosanelli (changes requested):
there's a layout problem in modules index when some custom filters are defined 

New version
image


Previous version (however not optimal)
image

src/Template/Pages/Dashboard/index.twig
qwerg: Ok, I agree on a specific `Categories` view for non-admin, and keeping both anchors commented for now.

stefanorosanelli (approved):

PR #569, opened 2021-05-28
Colophon

didoda:
This provides a value for 'project.colophon' and uses it in colophon.twig

### DIFF
--- a/src/Controller/Component/ModulesComponent.php
+++ b/src/Controller/Component/ModulesComponent.php
@@ -184,7 +184,7 @@ class ModulesComponent extends Component
         $project = [
             'name' => (string)Configure::read('Project.name', Hash::get($meta, 'project.name')),
             'version' => Hash::get($meta, 'version', ''),
-            'colophon' => '', // TODO: populate this value.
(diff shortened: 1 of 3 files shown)


src/Controller/Component/ModulesComponent.php
```diff
@@ -184,7 +184,7 @@ public function getProject(): array
         $project = [
             'name' => (string)Configure::read('Project.name', Hash::get($meta, 'project.name')),
             'version' => Hash::get($meta, 'version', ''),
-            'colophon' => '', // TODO: populate this value.
+            'colophon' => 'LGPL <a href="https://www.bedita.com" alt="BEdita Web site" target="_blank">www.bedita.com</a>',
```
stefanorosanelli: Not completely sure about this: `project.colophon` could be something project specific to add (from project metadata), if set - but no such information is in the API for now.

I'd prefer to simply remove it for now and stick with current static `www.bedita.com` link

didoda: fair enough

PR #575, opened 2021-06-04
BEM-1: roles config for modules

didoda:
This introduces a `AccessControl` configuration to handle modules access by role.

Config example:
```
'AccessControl' => [
    'manager' => [
        'hidden' => [],
        'readonly' => [],
    ],
    'guest' => [
        'hidden' => ['users', 'profiles'],
        'readonly' => ['documents'],
    ],
],
```

Rif: https://chialab.atlassian.net/browse/BEM-1


config/app.default.php
```diff
@@ -468,6 +468,20 @@
         ],
     ],
 
+    /**
+     * Modules accesses per role(s)
+     */
+    // 'Roles' => [
```
stefanorosanelli: Can we use `AccessControl` as key instead of `Roles`? Or `Access` or `ACL`... 

[superminor] I would not use `admin` in these examples because it's not yet available and will probably be `admin`-role only by default

src/Controller/Component/ModulesComponent.php
```diff
@@ -145,10 +145,39 @@ function (&$data, $key) use ($metaModules) {
...
+     * Module can be "hidden": remove from $modules.
+     * Module can be "readonly": add "'readonly' => true" to module.
+     *
+     * @param array $modules The modules
+     * @return array
+     */
+    protected function modulesByRoleConfig(array $modules): array
+    {
+        $user = $this->getController()->Auth->user();
+        $role = (string)Hash::get($user, 'roles.0');
```
stefanorosanelli: We should check against all roles not just the first one. 
We could limit to a list of roles via an `array_intersect()` of user roles and roles defined in configuration.

src/Controller/Component/ModulesComponent.php
```diff
@@ -145,10 +145,39 @@ function (&$data, $key) use ($metaModules) {
...
+        $hidden = (array)Configure::read(sprintf('Roles.%s.hidden', $role));
+        $readonly = (array)Configure::read(sprintf('Roles.%s.readonly', $role));
+        $keys = array_keys($modules);
+        foreach ($keys as $moduleName) {
+            if (in_array($moduleName, $hidden)) {
+                unset($modules[$moduleName]);
+                continue;
+            }
+            if (in_array($moduleName, $readonly)) {
+                $modules[$moduleName]['readonly'] = true;
```
stefanorosanelli: it's ok, but we could set `hints.allow` array to `array_diff($allow, ['POST', 'PATCH', 'DELETE'])` and don't change `PermsHelper::isAllowed()`

src/Controller/Component/ModulesComponent.php
```diff
@@ -145,10 +145,39 @@ function (&$data, $key) use ($metaModules) {
...
+     * @param array $modules The modules
+     * @return array
+     */
+    protected function modulesByRoleConfig(array $modules): array
+    {
+        $user = $this->getController()->Auth->user();
+        $role = (string)Hash::get($user, 'roles.0');
+        $hidden = (array)Configure::read(sprintf('Roles.%s.hidden', $role));
+        $readonly = (array)Configure::read(sprintf('Roles.%s.readonly', $role));
+        $keys = array_keys($modules);
```
stefanorosanelli: To remove hidden modules:
```php
$modules = array_diff_key($modules, array_flip($hidden));
```

stefanorosanelli (changes requested):

src/Controller/Component/ModulesComponent.php
didoda: `AccessControl` is already a configuration.
I would avoid too much configuration... `AccessControl` configuration keys could do the task: providing user roles

src/Controller/Component/ModulesComponent.php
didoda: in `modulesFromMeta`? not clear how to do that without changing something in `PermsHelper::isAllowed`.

src/Controller/Component/ModulesComponent.php
```diff
@@ -145,10 +145,40 @@ function (&$data, $key) use ($metaModules) {
...
+     * @return array
+     */
+    protected function modulesByRoleConfig(array $modules): array
+    {
+        $user = $this->getController()->Auth->user();
+        $roles = (array)Hash::get($user, 'roles');
+        $hidden = [];
+        $readonly = [];
+        foreach ($roles as $role) {
+            $hidden = array_merge($hidden, (array)Configure::read(sprintf('AccessControl.%s.hidden', $role)));
```
stefanorosanelli: not sure about `array_merge` => if a module is `hidden` on a role but not in another I think it should not be hidden

src/Controller/Component/ModulesComponent.php
```diff
@@ -145,10 +145,40 @@ function (&$data, $key) use ($metaModules) {
...
+            $readonly = array_merge($readonly, (array)Configure::read(sprintf('AccessControl.%s.readonly', $role)));
+        }
```
stefanorosanelli, line 175: we could `return $modules` if `$hidden` and `$readonly` are both empty

src/Controller/Component/ModulesComponent.php
stefanorosanelli: I suggest something like this

```php
// make sure $readonly contains valid module names
$readonly = array_intersect($readonly, array_keys($modules));
foreach ($readonly as $key) {
   $path = sprintf('%s.hints.allow', $key);
   $allow = (array)Hash::get($modules, $path);
   Hash::insert($modules, $path, array_diff($allow, ['POST', 'PATCH', 'DELETE']));
}
```
This way no change in `PermsHelper` is necessary since perms logic uses `hints.allow` only

src/Controller/Component/ModulesComponent.php
```diff
@@ -145,10 +145,40 @@ function (&$data, $key) use ($metaModules) {
             array_diff_key($metaModules, $modules),
             $pluginModules
         );
+        $this->modules = $this->modulesByRoleConfig($this->modules);
```
stefanorosanelli: we could use `$this->modules` instead of passing it by argument

```php
$this->modulesByRoleConfig();
```
Inside `modulesByRoleConfig()` change `$this->modules` directly

stefanorosanelli (approved):

PR #581, opened 2021-07-16
Concurrent editors in object's view page

didoda:
This provides a visible "Alert! Concurrent users: ..." in object view, when more than one user accesses an object view page.

How long a user is considered "active editor" on an object's view page? this is configurable with:

```
/**
 * Editors configuration.
 * concurrentCheckTime: the time in milliseconds that a concurrent access is considered still active
 */
'Editors' => [
     'concurrentCheckTime' => 30000, // 30 seconds
],
```

Default `concurrentCheckTime` is 20000 (20 seconds).

The system cleans up concurrent editors data, on each page access; expired users accesses are removed.

Note: we used `Cache` to save the `objects_editors` data: it is multi-user and it doesn't need database.

Rif: https://chialab.atlassian.net/browse/BEM-26

### DIFF
--- a/src/Controller/Component/ObjectsEditorsComponent.php
+++ b/src/Controller/Component/ObjectsEditorsComponent.php
@@ -1,0 +1,131 @@
+<?php
+namespace App\Controller\Component;
+
+use Cake\Cache\Cache;
+use Cake\Controller\Component;
+use Cake\Core\Configure;
+
+/**
+ * ObjectsEditors component
+ */
+class ObjectsEditorsComponent extends Component
+{
+    /**
+     * Objects editors.
+     *
+     * @var array
+     */
+    public $objectsEditors;
+
+    /**
+     * Concurrent check time.
+     *
+     * @var integer
+     */
+    public $concurrentCheckTime = 20000;
+
+    /**
+     * {@inheritDoc}
+     */
+    public function initialize(array $config): void
+    {
+        // set concurrent check time by config, if any. default 20s
+        $concurrentCheckTime = Configure::read('Editors.concurrentCheckTime');
+        if (!empty($concurrentCheckTime)) {
+            $this->concurrentCheckTime = $concurrentCheckTime;
+        }
+        // get concurrent editors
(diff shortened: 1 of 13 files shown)


src/Controller/Component/ObjectsEditorsComponent.php
```diff
@@ -0,0 +1,138 @@
...
+     * @return string|null
+     */
+    public function editorName(): ?string
+    {
+        $user = $this->getController()->Auth->user();
+        if (empty($user)) {
+            return null;
+        }
+
+        if (!empty($user['attributes']['name']) && !empty($user['attributes']['name'])) {
```
le0m:
```suggestion
        if (!empty($user['attributes']['name']) && !empty($user['attributes']['surname'])) {
```

src/Controller/Component/ObjectsEditorsComponent.php
```diff
@@ -0,0 +1,138 @@
...
+                $found = true;
+                $objectEditor['timestamp'] = $now;
+            }
+        }
+        if (!$found) {
+            $this->objectsEditors[$id][] = [
+                'name' => $editor,
+                'timestamp' => $now,
+            ];
+        }
```
le0m: I suggest either:
```suggestion
        foreach ($this->objectsEditors[$id] as &$objectEditor) {
            if ($objectEditor['name'] === $editor) {
                $found = true;
                $objectEditor['timestamp'] = $now;
                break;
            }
        }
        if (!$found) {
            $this->objectsEditors[$id][] = [
                'name' => $editor,
                'timestamp' => $now,
            ];
        }
```

or

```suggestion
        $editorKey = array_search($editor, array_column($this->objectsEditors[$id], 'name'));
        if ($editorKey === false) {
            $this->objectEditors[$id][] = [
                'name' => $editor,
                'timestamp' => $now,
            ];
        } else {
            $this->objectEditors[$id][$editorKey]['timestamp'] = $now;
        }
```

le0m (approved):

PR #585, opened 2021-07-22
refactor: rename methods and routes {action}Json to {action}

didoda:
After https://github.com/bedita/manager/pull/580 `ModulesController::save` was not used anymore.

This remove "old"  `ModulesController::save` and rename `saveJson` to `save`.

This renames routes and methods with `{action}Json` to `{action}`

Bonus: more coverage for `TranslationsControllerTest`.

### DIFF
--- a/src/Controller/ModulesController.php
+++ b/src/Controller/ModulesController.php
@@ -279,6 +279,9 @@ class ModulesController extends AppController
             $this->set(compact('error'));
             $this->set('_serialize', ['error']);
 
+            // set session data to recover form
+            $this->Modules->setDataFromFailedSave($this->objectType, $requestData);
+
             return;
         }
         if ($response['data']) {
--- a/src/Controller/ModulesController.php
+++ b/src/Controller/ModulesController.php
@@ -374,7 +374,7 @@ class ModulesController extends AppController
      * @param string $relation The relation name.
      * @return void
      */
-    public function relatedJson($id, string $relation): void
+    public function related($id, string $relation): void
     {
         if ($id === 'new') {
             $this->set('data', []);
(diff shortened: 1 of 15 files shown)


src/Template/Layout/js/app/pages/modules/view.js
```diff
@@ -48,7 +48,11 @@ export default {
             event.stopPropagation();
             const form = new FormData(event.target);
             const action = event.target.getAttribute('action');
-            const response = await fetch(`${action}Json`, {
+            let endpoint = action;
+            if (!endpoint.endsWith('/save')) {
+                endpoint = `${action}Json`;
+            }
+            const response = await fetch(`${endpoint}`, {
```
le0m: I think it's safe to just use the form's action (see [here](https://github.com/bedita/manager/blob/master/src/Template/Pages/Modules/view.twig#L21)) and remove this `if`. This Vue component is used only in the linked template, so there shouldn't be any side-effects

```suggestion
            const response = await fetch(action, {
```

le0m:
As there are other `...Json` functions in `ModulesController`, I'd prefer to have them named consistently, either all or none with the suffix `Json`

src/Controller/ModulesController.php
```diff
@@ -251,58 +251,12 @@ function ($schema) {
...
-            $response = $this->apiClient->save($this->objectType, $requestData);
-            $objectId = (string)Hash::get($response, 'data.id');
-            $this->Modules->saveRelated($objectId, $this->objectType, $relatedData);
-        } catch (InternalErrorException | BEditaClientException | UploadException $e) {
-            // Error! Back to object view or index.
-            $this->log($e, LogLevel::ERROR);
-            $this->Flash->error($e->getMessage(), ['params' => $e]);
-
-            // set session data to recover form
-            $this->Modules->setDataFromFailedSave($this->objectType, $requestData);
```
stefanorosanelli, line 281: I think we should move `$this->Modules->setDataFromFailedSave($this->objectType, $requestData);` in the new `save()`

le0m (approved):

stefanorosanelli (approved):

PR #587, opened 2021-08-02
fix: render flash messages even when using AJAX save

le0m:
This PR adds rendering of flash messages when saving an object with ajax request returns an error:
- action in `DashboardController` to return the template for flash messages
- element to render only flash messages template
- method in `view.js` to dynamically render flash messages with a child Vue instance

### DIFF
--- a/src/Controller/ModulesController.php
+++ b/src/Controller/ModulesController.php
@@ -320,10 +320,10 @@ class ModulesController extends AppController
             $objectId = (string)Hash::get($response, 'data.id');
             $this->Modules->saveRelated($objectId, $this->objectType, $relatedData);
         } catch (BEditaClientException $error) {
-            $this->log($error, LogLevel::ERROR);
+            $this->log($error->getMessage(), LogLevel::ERROR);
             $this->Flash->error($error->getMessage(), ['params' => $error]);
 
-            $this->set(['error' => $error->getMessage()]);
+            $this->set(['error' => $error->getAttributes()]);
             $this->set('_serialize', ['error']);
 
             return;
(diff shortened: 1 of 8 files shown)


src/Controller/ModulesController.php
```diff
@@ -321,8 +321,9 @@ public function saveJson(): void
             $this->Modules->saveRelated($objectId, $this->objectType, $relatedData);
         } catch (BEditaClientException $error) {
             $this->log($error, LogLevel::ERROR);
```
stefanorosanelli: Currently the `error` object received in the Vue app is empty. I suggest this change in order to get a meaningful response and avoid the subsequent `/dashboard/messages` ajax call that can be avoided 


```php
} catch (BEditaClientException $ex) {
            // this has the same effect as the current log
            $this->log($ex->getMessage(), LogLevel::ERROR);
            // meaningful messages received by API (i.e. `error` response is available in exception attributes)
            // if `error` is empty a generic error message can be displayed
            $this->set(['error' => $ex->getAttributes()]);
            $this->set('_serialize', ['error']);

            return;

```

src/Template/Layout/js/app/pages/modules/view.js
```diff
@@ -97,6 +98,28 @@ export default {
             document.body.appendChild(iframe.$el);
         },
 
+        async showFlashMessages() {
+            // fetch flash messages template
+            const messages = await fetch('/dashboard/messages');
```
stefanorosanelli, line 103: with the suggested change above there's no need to call `'/dashboard/messages'`
an error object can be passed to this method and displayed

didoda (approved):

PR #590, opened 2021-08-26
Translations update

didoda:
This fixes and updates some translations.

 - get a translation from plugin, for modelling labels, if available (new method `Layout.typeLabel`)
 - refactor javascript files to properly use `t` method, otherwise ttag extract won't extract strings
 - update po files, for italian translations

### DIFF
--- a/src/View/Helper/LayoutHelper.php
+++ b/src/View/Helper/LayoutHelper.php
@@ -134,12 +134,13 @@ class LayoutHelper extends Helper
      */
     public function typeLabel(string $type): ?string
     {
-        $plugins = Configure::read('Plugins');
-        if (empty($plugins)) {
-            return __($type);
+        $res = __($type);
+        $pluginName = Hash::get((array)Configure::read('Plugins'), '0');
+        // if we have no actual translation and a plugin let's try with plugin's gettext
+        if ($pluginName && $res === $type) {
+           return __d($pluginName, $type);
         }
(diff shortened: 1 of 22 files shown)


src/View/Helper/LayoutHelper.php
```diff
@@ -125,4 +125,21 @@ public function customElement(string $item, string $type = 'relation'): string
...
+     */
+    public function typeLabel(string $type): ?string
+    {
+        $plugins = Configure::read('Plugins');
+        if (empty($plugins)) {
+            return __($type);
+        }
+        $pluginName = array_keys($plugins)[0];
+
+        return __d($pluginName, $type);
```
stefanorosanelli: Is this looking only in plugin's gettext? 
If so it will be better to look in core gettext files first, otherwise we could miss translations in `core` - something like: 

```php 
    $res = __($type); 
    $pluginName = Hash::get((array)Configure::read('Plugins'), '0'); 
    // if we have no actual translation and a plugin let's try with plugin's gettext
    if ($pluginName && $res === $type) {
       return __d($pluginName, $type);
    }

    return $res;
```

stefanorosanelli (approved):

src/View/Helper/LayoutHelper.php
didoda: fair enough. I'll change this part, as you suggested

PR #592, opened 2021-08-27
BEM-36: Add my contents checkbox

paolosantarsiero:
Add a checkbox in modules views to filter items created and modified for last time by current user.


src/Template/Layout/js/app/app.js
```diff
@@ -176,6 +176,11 @@ const _vueInstance = new Vue({
          * @return {void}
          */
         onFilterObjects(filter) {
+            // remove from query string ilter mine if it is false
+            if (!filter.filter.mine) {
```
stefanorosanelli: why do we need this?

stefanorosanelli (commented):
Seems ok

On default filters the checkbox is too close to the `Reset Filters` button, see below

image

src/Template/Layout/js/app/app.js
paolosantarsiero: I need this to remove `filter.mine` from `url` (in query string) when it is false and unchecked

paolosantarsiero: This happened because when `filter.mine` is `false` or `0` it seems like enabled.

stefanorosanelli (approved):

stefanorosanelli:
currently we have only `mine` filter => objects created by the logged user
it's fine for me now

didoda (approved):

didoda (approved):

PR #595, opened 2021-08-30
Load users info in metadata of objects

paolosantarsiero:
Add a function to load user info in metadata properties. Replace user id with:

- `name surname` when they are not undefined
- `username` else

### DIFF
--- a/src/Template/Element/Form/meta.twig
+++ b/src/Template/Element/Form/meta.twig
@@ -10,7 +10,7 @@
             <h2>{{ __('Metadata') }}</h2>
         </header>
 
-        <div v-show="isOpen" class="tab-container">
+        <div v-show="isOpen && !isLoading" class="tab-container">
             <table>
                 <tbody>
                     <tr>
--- a/src/Template/Element/Form/meta.twig
+++ b/src/Template/Element/Form/meta.twig
@@ -1,7 +1,7 @@
 {% set meta = (object) ? object.meta : resource.meta %}
 
 {% if meta %}
-<property-view inline-template :tab-open="tabsOpen" tab-name="meta">
+<property-view inline-template :tab-open="tabsOpen" tab-name="meta" :object="object">
     <section class="fieldset">
 
         <header @click.prevent="toggleVisibility()"
@@ -20,7 +20,7 @@
                     {% for key, val in meta %}
                         <tr>
                             <th>{{ key }}:</th>
-                            <td>{{ Schema.format(val, schema.properties[key]) }}</td>
+                            <td id="{{ key }}">{{ Schema.format(val, schema.properties[key]) }}</td>
                         </tr>
                     {% endfor %}
                 </tbody>
--- a/src/Template/Element/Form/meta.twig
+++ b/src/Template/Element/Form/meta.twig
@@ -20,7 +20,7 @@
                     {% for key, val in meta %}
                         <tr>
                             <th>{{ key }}:</th>
-                            <td id="{{ key }}">{{ Schema.format(val, schema.properties[key]) }}</td>
+                            <td name="{{ key }}">{{ Schema.format(val, schema.properties[key]) }}</td>
                         </tr>
                     {% endfor %}
                 </tbody>
--- a/src/Template/Layout/js/app/components/property-view/property-view.js
+++ b/src/Template/Layout/js/app/components/property-view/property-view.js
@@ -20,7 +20,7 @@
         'accept': 'application/json',
     }
 };
- 
+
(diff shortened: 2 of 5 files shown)


src/Template/Element/Form/meta.twig
```diff
@@ -20,7 +20,7 @@
                     {% for key, val in meta %}
                         <tr>
                             <th>{{ key }}:</th>
-                            <td>{{ Schema.format(val, schema.properties[key]) }}</td>
+                            <td id="{{ key }}">{{ Schema.format(val, schema.properties[key]) }}</td>
```
didoda: it's not safe to have multiple td with same id. maybe `name` or other attribute could be better, in this context

src/Template/Layout/js/app/components/property-view/property-view.js
```diff
@@ -131,5 +149,28 @@ export default {
...
+            const userJson = await userRes.json();
+            const users = userJson.data;
+
+            users.map((user) => {
+                const href = `${BEDITA.base}/${user.type}/view/${user.id}`;
+                const userInfo = (user.attributes.name  != undefined || user.attributes.surname != undefined) 
+                        ? user.attributes.name + ' ' + user.attributes.surname
+                        : user.attributes.uname;
+
+                if(user.id == creatorId) {
```
didoda:
```suggestion
                if (user.id === creatorId) {
```
chore fix: space and operator

src/Template/Layout/js/app/components/property-view/property-view.js
```diff
@@ -131,5 +149,28 @@ export default {
...
+                    document.querySelector(`td[id='created_by']`).innerHTML = `<a href="${href}">${userInfo}</a>`;
+                }
+                
+                if (user.id == modifierId) {
```
didoda:
```suggestion
                if (user.id === modifierId) {
```
chore fix: operator

src/Template/Layout/js/app/components/property-view/property-view.js
```diff
@@ -131,5 +149,28 @@ export default {
                 this.$refs.relation.loadRelatedObjects();
             }
         },
+        async loadInfoUsers() {
+            const creatorId = this.object?.meta?.created_by;
+            const modifierId = this.object?.meta?.modified_by;
+            let usersId = [creatorId, modifierId];
```
didoda:
```suggestion
            const usersId = [creatorId, modifierId];
```
chore fix: use const

src/Template/Layout/js/app/components/property-view/property-view.js
paolosantarsiero: `user.id` is a String and `creatorId` is a Integer. Do you suggest to cast `user.id` in a integer and after apply this operator?

didoda: you can use `==`, fine. maybe an inline comment, to explain why, as `// user.id is string, creatorId is int

src/Template/Layout/js/app/components/property-view/property-view.js
```diff
@@ -131,5 +149,29 @@ export default {
...
+            const usersId = [creatorId, modifierId];
+            const userRes = await fetch(`${API_URL}api/users?filter[id]=${usersId.join(',')}`, API_OPTIONS);
+            const userJson = await userRes.json();
+            const users = userJson.data;
+
+            users.map((user) => {
+                const href = `${BEDITA.base}/view/${user.id}`;
+                const userInfo = (user.attributes.name  != undefined || user.attributes.surname != undefined) 
+                        ? user.attributes.name + ' ' + user.attributes.surname
+                        : user.attributes.uname;
```
stefanorosanelli: `username` instead of `uname`

src/Template/Layout/js/app/components/property-view/property-view.js
```diff
@@ -131,5 +149,29 @@ export default {
                 this.$refs.relation.loadRelatedObjects();
             }
         },
+        async loadInfoUsers() {
+            const creatorId = this.object?.meta?.created_by;
+            const modifierId = this.object?.meta?.modified_by;
+            const usersId = [creatorId, modifierId];
+            const userRes = await fetch(`${API_URL}api/users?filter[id]=${usersId.join(',')}`, API_OPTIONS);
```
stefanorosanelli: [minor] I'd add this query string
```
fields[users]=name,surname,username
```
to load only those fields

stefanorosanelli (approved):

didoda (approved):

PR #597, opened 2021-08-30
Object lock and unlock

didoda:
This provides lock and unlock on objects.

Rif: https://chialab.atlassian.net/browse/BEM-24


src/Controller/LockController.php
```diff
@@ -0,0 +1,59 @@
...
+     */
+    protected function lock(bool $locked): void
+    {
+        $type = $this->request->getParam('object_type');
+        $id = $this->request->getParam('id');
+        $meta = compact('locked');
+        $data = compact('id', 'type', 'meta');
+        $payload = json_encode(compact('data'));
+        try {
+            $this->apiClient = ApiClientProvider::getApiClient();
```
stefanorosanelli: Not necessary - `$this->apiClient` already instantiated by `AppController`

src/Controller/LockController.php
```diff
@@ -0,0 +1,59 @@
+<?php
+namespace App\Controller;
```
stefanorosanelli, line 13: missing copyright header

src/Template/Pages/Modules/view.twig
```diff
@@ -25,6 +25,7 @@
             })|raw }}
 
             {{ Form.hidden('id', {'value': (object) ? object.id : resource.id})|raw }}
+            {{ Form.hidden('locked', {'value': (object) ? object.meta.locked : false})|raw }}
```
stefanorosanelli: are we using this `locked` form field anywhere? can we remove it?

src/Template/Pages/Modules/view.twig
```diff
@@ -115,8 +116,26 @@
...
 
+            {# Append "Lock/Unlock" #}
+            {% if object.id and Perms.canLock() %}
+                {% if object.meta.locked %}
+                    {% do _view.append('module-buttons', Html.link(
+                        __('Unlock'),
+                        {'_name': 'lock:remove', 'object_type': objectType, 'id': object.id},
+                        {'class': 'button button-outlined button-outlined-hover-module-' ~ currentModule.name }
+                    )) %}
+                    {{ Form.hidden('locked', {'value': object.meta.locked})|raw }}
```
stefanorosanelli: are we using this `locked` form field anywhere?

src/Controller/LockController.php
```diff
@@ -0,0 +1,59 @@
...
+{
+    /**
+     * Add lock
+     *
+     * @return void
+     */
+    public function add(): void
+    {
+        $this->lock(true);
+        $this->redirect($this->referer());
```
stefanorosanelli: the proper way to redirect is to use 
```php 
  return $this->redirect(..);
``` 
same below

src/Template/Element/Modules/index_table_row.twig
```diff
@@ -89,6 +91,8 @@
                         'class': buttonClasses,
                     }
                 )|raw}}
+            {% elseif object.meta.locked %}
+                [ {{ __('Locked') }} ]
```
stefanorosanelli: I'd leave it empty for now - or use a `locked` icon

src/Controller/LockController.php
```diff
@@ -0,0 +1,59 @@
...
+        $payload = json_encode(compact('data'));
+        try {
+            $this->apiClient = ApiClientProvider::getApiClient();
+            $this->apiClient->patch(
+                sprintf('/%s/%s', $type, $id),
+                $payload,
+                ['Content-Type' => 'application/json']
+            );
+        } catch (BEditaClientException $ex) {
+            $this->loadComponent('App.Flash', ['clear' => true]);
```
stefanorosanelli: maybe an additional error log?

`$this->loadComponent()` not necessary, already provided by `AppController`

src/Controller/AppController.php
```diff
@@ -244,6 +244,14 @@ function ($item) {
                 return ['name' => $category];
             }, $data['categories']);
         }
+
+        // locked
+        $locked = (string)Hash::get($data, 'locked');
```
stefanorosanelli: Do we need this? Is `locked` form field necessary?
(see also comments below)

stefanorosanelli (changes requested):

src/Template/Pages/Modules/view.twig
didoda: yes, it's used to remove "status" and "uname" from payload, on data save, when object is locked

src/Controller/AppController.php
didoda: this provides the server side logic, to remove from payload "status" and "uname", when saving a locked object. not necessary, but useful IMHO

stefanorosanelli (approved):

PR #599, opened 2021-09-06
Translations module

didoda:
This provides a basic translations module.

 - Pagination: yes
 - Filters: yes, with [PR on BEdita Core](https://github.com/bedita/bedita/pull/1818)
 - Bulk: no

Rif: https://chialab.atlassian.net/browse/BEM-42

### DIFF
--- a/src/Controller/TranslationsController.php
+++ b/src/Controller/TranslationsController.php
@@ -44,6 +44,7 @@ class TranslationsController extends ModulesController
     public function add($id): ?Response
     {
         $this->request->allowMethod(['get']);
+        $this->objectType = substr($this->request->here, 1, strpos(substr($this->request->here, 1), '/'));
 
         try {
             $response = $this->apiClient->getObject($id, $this->objectType);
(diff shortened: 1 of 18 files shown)


src/Controller/TranslationsController.php
```diff
@@ -25,6 +25,32 @@
  */
 class TranslationsController extends ModulesController
 {
+    /**
+     * @inheritDoc
+     */
+    public function index(): ?Response
```
stefanorosanelli: I suggest a different approach, in `TranslationsController` we could have

```php

    public function initialize(): void
    {
        $this->request->setParam('object_type', 'translations'); 
        parent::initialize();
        $this->Query->setConfig('include', 'objects'); 
     }
```
Then in `QueryComponent` we should add `include=###` to query string in `index()` if the `include` config is set.
This way we don't need any `TranslationsController::index()` method.
Translated objects will be embedded on every resource in the response in `relationships.object.data` 

Then, in a future PR, a refactor may be necessary in order to define a `Modules` Controller class that can be extended by modules that not represent objects (like `translations`) and an `ObjectModules` controller used and extended by modules representing object types.
For instance we should have a `resource_type` instead of `object_type` to be more general.
Also a twig related refactor could be useful.

PR #607, opened 2021-09-28
Enhance bulk actions

didoda:
This introduces more bulk actions, and a refactor of "change attribute" bulk action.

In index page, bulk actions available are:

 - export bulk / export all (by filter)
 - n "attribute change" bulk (for each config `Properties.<module>.bulk` item)
 - "category" bulk
 - "position" bulk
 - trash bulk

New features:

 - category picker (based on https://vue-treeselect.js.org): choose multiple categories to associate to an object
 - folder picker (based on https://vue-treeselect.js.org): choose one folder as destination for "move to" or "copy to"

Bonus: more tests for SchemaComponent and ModulesController (coverage 100%).

# a preview

bulk_modello

Rif: https://chialab.atlassian.net/browse/BEM-30

### DIFF
--- a/src/Controller/BulkController.php
+++ b/src/Controller/BulkController.php
@@ -40,7 +40,7 @@ class BulkController extends AppController
     /**
      * Selected categories
      *
-     * @var array
+     * @var array|string
      */
     protected $categories = [];
 
--- a/src/Controller/BulkController.php
+++ b/src/Controller/BulkController.php
@@ -141,7 +141,7 @@ class BulkController extends AppController
      */
     protected function loadCategories(): void
     {
-        $url = sprintf('/model/categories?filter[type]=%s&filter[id]=%s', $this->objectType, $this->categories);
+        $url = sprintf('/model/categories?filter[type]=%s&filter[id]=%s', $this->objectType, (string)$this->categories);
         try {
             $response = $this->apiClient->get($url, ['page_size' => 100]);
             $this->categories = (array)Hash::extract($response, 'data.{n}.attributes.name');
@@ -161,12 +161,12 @@ class BulkController extends AppController
             try {
                 $object = $this->apiClient->getObject($id, $this->objectType);
                 $objectCategories = (array)Hash::extract($object, 'data.attributes.categories.{n}.name');
-                $this->categories = !empty($objectCategories) ? array_unique(array_merge($this->categories, $objectCategories)) : $this->categories;
+                $this->categories = !empty($objectCategories) ? array_unique(array_merge((array)$this->categories, $objectCategories)) : $this->categories;
                 $payload = compact('id');
                 $payload['categories'] = array_map(function ($category) {
 
                     return ['name' => $category];
-                }, $this->categories);
+                }, (array)$this->categories);
                 $this->apiClient->save($this->objectType, $payload);
             } catch (BEditaClientException $e) {
                 $this->errors[] = ['id' => $id, 'message' => $e->getAttributes()];
--- a/src/Controller/BulkController.php
+++ b/src/Controller/BulkController.php
@@ -1,0 +1,235 @@
+<?php
+/**
+ * BEdita, API-first content management framework
+ * Copyright 2021 ChannelWeb Srl, Chialab Srl
+ *
+ * This file is part of BEdita: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * See LICENSE.LGPL or <http://gnu.org/licenses/lgpl-3.0.html> for more details.
+ */
+namespace App\Controller;
+
+use App\Controller\AppController;
+use BEdita\SDK\BEditaClientException;
+use Cake\Http\Response;
+use Cake\Utility\Hash;
+use Psr\Log\LogLevel;
+
+/**
+ * Bulk Controller.
+ */
+class BulkController extends AppController
+{
+    /**
+     * Object type currently used
+     *
+     * @var string
+     */
+    protected $objectType = null;
+
+    /**
+     * Selected objects IDs
+     *
+     * @var array
+     */
+    protected $ids = [];
+
+    /**
+     * Selected categories
+     *
+     * @var array
+     */
+    protected $categories = [];
+
+    /**
+     * Errors
+     *
+     * @var array
+     */
+    protected $errors = [];
+
+    /**
+     * {@inheritDoc}
+     */
+    public function initialize(): void
+    {
+        parent::initialize();
+
+        $this->objectType = $this->request->getParam('object_type');
+    }
+
+    /**
+     * Bulk change attribute value for selected ids.
+     *
+     * @return \Cake\Http\Response|null
+     */
+    public function attribute(): ?Response
+    {
+        $this->request->allowMethod(['post']);
+        $requestData = $this->request->getData();
+        $this->ids = explode(',', (string)Hash::get($requestData, 'ids'));
+        $this->saveAttribute($requestData['attributes']);
+        $this->errors();
+
+        return $this->redirect(['_name' => 'modules:list', 'object_type' => $this->objectType, '?' => $this->request->getQuery()]);
+    }
+
+    /**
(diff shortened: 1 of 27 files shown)


src/Controller/BulkController.php
```diff
@@ -0,0 +1,249 @@
...
+     * Load categories by type and ids.
+     *
+     * @return void
+     */
+    protected function loadCategories(): void
+    {
+        $url = sprintf('/model/categories?filter[type]=%s&filter[id]=%s', $this->objectType, (string)$this->categories);
+        try {
+            $response = $this->apiClient->get($url, ['page_size' => 100]);
+            $this->categories = (array)Hash::extract($response, 'data.{n}.attributes.name');
```
stefanorosanelli: Can we used `categories` already present in `schema` cache? 

For example using 
```php
$schema = $this->Schema->getSchema($this->objectType);
```

src/Controller/BulkController.php
```diff
@@ -0,0 +1,249 @@
...
+    /**
+     * Save categories for selected objects.
+     *
+     * @return void
+     */
+    protected function saveCategories(): void
+    {
+        foreach ($this->ids as $id) {
+            try {
+                $object = $this->apiClient->getObject($id, $this->objectType);
```
stefanorosanelli: [minor] retrieve only categories 
```php
$this->apiClient->getObject($id, $this->objectType, ['fields' => 'categories']);
```

src/Controller/BulkController.php
```diff
@@ -0,0 +1,249 @@
...
+                $objectCategories = (array)Hash::extract($object, 'data.attributes.categories.{n}.name');
+                $this->categories = !empty($objectCategories) ? array_unique(array_merge((array)$this->categories, $objectCategories)) : $this->categories;
```
stefanorosanelli: [minor] it works even on empty `$objectCategories`

```php
$this->categories = array_unique(array_merge((array)$this->categories, $objectCategories));
```

src/Controller/BulkController.php
```diff
@@ -0,0 +1,249 @@
...
+     * Handle page errors.
+     *
+     * @return void
+     */
+    protected function errors(): void
+    {
+        if (empty($this->errors)) {
+            return;
+        }
+        $this->log($this->errors, LogLevel::ERROR);
```
stefanorosanelli: Not sure if  `$this->log()` with array as message works as expected... :thinking:

src/Template/Element/Modules/index_bulk.twig
```diff
@@ -1,92 +1,17 @@
 <div class="bulk-actions" :class="!selectedRows.length? '' : 'enabled'">
+    <nav>
+        {% element 'Form/bulk_export' %}
+    </nav>
     <header>
-        <p>{{ __('Actions on selected items') }}</p>
+        <h1>{{ __('Actions on selected items') }}</h1>
```
stefanorosanelli: [superminor] use `<h2>` or `<h3>` maybe?

src/Template/Layout/js/app/components/folder-picker/folder-picker.js
```diff
@@ -0,0 +1,72 @@
...
+
+    methods: {
+
+        async fetchFolders(parent) {
+            const folders = [];
+            let filter = !parent ? 'filter[roots]' : `filter[parent]=${parent.id}`;
+            let page = 1;
+            let i = 0;
+            do {
+                let response = await fetch(`${API_URL}api/folders?${filter}&page=${page}&page_size=100`, API_OPTIONS);
```
stefanorosanelli: It seems we are loading `/folders?filter[roots]` always on every index page call
Can we call it only on filter usage? Only when we are using `Copy to/Move to` functionality... 

Calls like `/folders?filter[parent]={id}` are already made upon request....

didoda: We should instantiate the picker only "on demand". Not easy, now, with the current UI. I'll take a look, however

src/Controller/BulkController.php
didoda: ok getting categories from schema; but we have to filter them by ids (matching the categories selected by user before submit).
something like
```
$schema = $this->Schema->getSchema($this->objectType);
$schemaCategories = (array)Hash::extract($schema, 'categories');
$ids = explode(',', $this->categories);
$this->categories = [];
foreach ($schemaCategories as $schemaCategory) {
    if (in_array($schemaCategory['id'], $ids)) {
        $this->categories[] = $schemaCategory['name'];
    }
}
```

src/Template/Layout/js/app/components/folder-picker/folder-picker.js
didoda: done... when choosing an action (move/copy), the folder roots are fetch.

stefanorosanelli (approved):
Ok as functionality, works as expected
Some UI refinements needed before merge

src/Template/Layout/js/app/components/category-picker/category-picker.js
```diff
@@ -0,0 +1,44 @@
...
+
+    mounted() {
+        this.loadCategories();
+    },
+
+    methods: {
+        loadCategories() {
+            this.categoriesOptions = [];
+            let i = 0;
+            this.categories.forEach(category => {
```
DarioSacco: you can use the `map` function to do the same thing in one line

```js
this.categoriesOptions = this.categories.map((category) => ({ id: category.id, label: category.label }));
```

src/Template/Layout/js/app/components/folder-picker/folder-picker.js
```diff
@@ -0,0 +1,83 @@
...
+                let json = await response.json();
+                if (json.data) {
```
DarioSacco: you can avoid the `i` variable also here

```js
if (json.data) {
  const newFolders = json.data.map((folder) => ({
    id: folder.id,
    label: folder.attributes.label,
    children: null,
  }));
  folders.push(...newFolders);
}
```

src/Template/Pages/Modules/_modules-index.scss
```diff
@@ -146,23 +146,41 @@ body.view-module {
...
+                            width: auto;
+                            color: #000;
+                        }
+                        .vue-treeselect__menu-container {
+                            color: #000;
+                        }
+                        .vue-treeselect__input-container {
+                            color: #000;
+                        }
+                        .vue-treeselect__placeholder {
```
DarioSacco, line 213: this `.vue-treeselect` rules can be collapsed

```scss
.vue-treeselect__control {
  width: auto;
}

.vue-treeselect__control,
.vue-treeselect__menu-container,
.vue-treeselect__input-container,
.vue-treeselect__placeholder {
  color: #000;
}
```

src/Template/Layout/js/app/components/folder-picker/folder-picker.js
```diff
@@ -0,0 +1,83 @@
...
+    },
+
+    methods: {
+
+        async fetchFolders(parent) {
+            const folders = [];
+            let filter = !parent ? 'filter[roots]' : `filter[parent]=${parent.id}`;
+            let page = 1;
+            let i = 0;
+            do {
```
DarioSacco: maybe calls after the first one can be parallelized (without the `await`)? the pagination infos can be fetched with the first call, then the while doesn't need the `true` condition but can be something like
```js
while(page <= paging.page_count)
```

DarioSacco (changes requested):

src/Template/Layout/js/app/components/folder-picker/folder-picker.js
didoda:
```
let response = await fetch(`${API_URL}api/folders?${filter}&page=${page}&page_size=100`, API_OPTIONS);
```
"more than 1 page to fetch" is a remote case. with "limit 100" more than 100 children are fetched. pagination is there, just in case...

DarioSacco (approved):

src/Template/Layout/js/app/components/folder-picker/folder-picker.js
didoda: a second thought: better a refactor to use the pattern suggested. done ;)

PR #612, opened 2021-10-08
Enhance index filters

DarioSacco:
Some new filters and enhancements in modules index page:
- "More filters" toggle
- filter by categories
- filter by position, with additional "descendants" switch to change between api filters `parent` and `ancestor`
- multiple status filter
- filters initial values

Other things:
- changed to `false` the default value of the "menu" toggle in view's `Position` section

this filters can be enabled by object properties configuration, setting the right values for the property `filter`.
E.g.

```php
'Properties' => [
  'events' => [
    'filter' => [
      'status',
      'lang',
      'categories'
      'parent'
    ]
  ]
]
```

### DIFF
--- a/src/Controller/Component/PropertiesComponent.php
+++ b/src/Controller/Component/PropertiesComponent.php
@@ -60,9 +60,7 @@ class PropertiesComponent extends Component
         ],
 
         'filter' => [
-            'status',
-            'lang',
-            'categories'
+            'status'
         ],
         'bulk' => [
             'status',
--- a/src/Controller/Component/PropertiesComponent.php
+++ b/src/Controller/Component/PropertiesComponent.php
@@ -60,7 +60,7 @@ class PropertiesComponent extends Component
         ],
 
         'filter' => [
-            'status'
+            'status',
         ],
         'bulk' => [
             'status',
--- a/src/Controller/Component/PropertiesComponent.php
+++ b/src/Controller/Component/PropertiesComponent.php
@@ -61,6 +61,8 @@ class PropertiesComponent extends Component
 
         'filter' => [
             'status',
+            'lang',
+            'categories'
         ],
         'bulk' => [
             'status',
--- a/src/Form/Options.php
+++ b/src/Form/Options.php
@@ -58,14 +58,15 @@ class Options
 
     /**
      * Options for lang, using configuration loaded from API
-     * If available, use config `I18n.languages`
+     * If available, use config `Project.config.I18n.languages`.
      *
      * @param mixed|null $value The field value.
      * @return array
      */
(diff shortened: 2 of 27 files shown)


src/Form/Options.php
```diff
@@ -58,13 +58,14 @@ public static function customControl(string $name, $value): array
...
      * Options for lang, using configuration loaded from API
-     * If available, use config `Project.config.I18n.languages`
+     * If available, use config `Project.config.I18n.languages`.
      *
      * @param mixed|null $value The field value.
      * @return array
      */
     public static function lang($value): array
     {
+        // $languages = Configure::read('I18n.languages');
```
didoda: This line can be removed

src/Template/Element/Form/bulk_export.twig
```diff
@@ -1,32 +1,29 @@
 {% if not currentModule.hints.multiple_types %}
-    {{ Form.create(null, {'id': 'form-export', 'class': 'bulk-actions', 'url': {'_name': 'export:export', 'object_type': objectType}})|raw }}
-
+    {{ Form.create(null, {
+        'id': 'form-export',
+        'url': {'_name': 'export:export', 'object_type': objectType}
+    })|raw }}
```
didoda, line 5: class `bulk-actions` is missing, compared to previous version. is it desired?

src/Template/Element/Form/bulk_trash.twig
```diff
@@ -1,21 +1,19 @@
...
-
-            </div>
-
-        </div>
-
-    {{ Form.end()|raw }}
-
+{{ Form.create(null, {
+    'id': 'form-delete',
+    'url': {'_name': 'modules:delete', 'object_type': objectType}
```
didoda, line 4: class `bulk-actions` missing here too

src/Controller/Component/PropertiesComponent.php
```diff
@@ -60,7 +60,7 @@ class PropertiesComponent extends Component
         ],
 
         'filter' => [
-            'status',
+            'status'
```
stefanorosanelli: Ending `,` is required by coding style

src/Template/Element/Form/bulk_export.twig
DarioSacco: yes. after some refactor in the style, that class was unnecessary

src/Template/Element/Form/bulk_trash.twig
DarioSacco: same here, I leaved that class only in the actions container

stefanorosanelli (approved):

PR #614, opened 2021-10-14
Fix filter in paginated-content

didoda:
This resolves a bug in filtering data by type in "add objects to relation" panel.

After https://github.com/bedita/manager/pull/612/ filter box contains more filters.

Bug reason was filter badly built (missing a `&` in the api call url).

### DIFF
--- a/src/Template/Layout/js/app/mixins/paginated-content.js
+++ b/src/Template/Layout/js/app/mixins/paginated-content.js
@@ -190,7 +190,7 @@ export const PaginatedContentMixin = {
                 const query = this.query[key];
 
                 if (key !== 'filter') {
-                    if (query) {
+                    if (query && typeof query !== 'object') {
                         urlSearchParams.append(key, query);
                     }
 
--- a/src/Template/Layout/js/app/mixins/paginated-content.js
+++ b/src/Template/Layout/js/app/mixins/paginated-content.js
@@ -189,28 +189,34 @@ export const PaginatedContentMixin = {
             Object.keys(this.query).forEach((key, index) => {
                 const query = this.query[key];
 
+                if (key !== 'filter') {
+                    if (query) {
+                        urlSearchParams.append(key, query);
+                    }
+
+                    return;
+                }
+
                 // parse filter property
-                if (key === 'filter') {
-                    Object.keys(query).forEach((filterKey) => {
-                        let filterVal = query[filterKey];
-                        if (filterVal) {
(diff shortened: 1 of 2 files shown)


src/Template/Layout/js/app/mixins/paginated-content.js
```diff
@@ -180,39 +180,41 @@ export const PaginatedContentMixin = {
...
+                                }
+                            } else {
+                                urlSearchParams.append(`filter[${filterKey}]`, filterVal);
+                            }
                         }
                     });
-
-                    entry = filter;
+                } else {
+                    urlSearchParams.append(key, query);
```
le0m: I would switch the conditions to return early:

```js
if (key !== 'filter') {
  urlSearchParams.append(key, query);

  return;
}

// here goes the code to handle key === 'filter'
```

src/Template/Layout/js/app/mixins/paginated-content.js
```diff
@@ -180,39 +180,41 @@ export const PaginatedContentMixin = {
...
 
                 // parse filter property
                 if (key === 'filter') {
-                    let filter = '';
                     Object.keys(query).forEach((filterKey) => {
-                        if (query[filterKey] !== '') {
-                            filter += `filter[${filterKey}]=${query[filterKey]}`;
+                        let filterVal = query[filterKey];
+                        if (filterVal) {
+                            if (typeof filterVal === 'object' || typeof filterVal === 'array') {
```
le0m: This won't work for objects, because they don't have `.length` nor `.join()` (used below)

didoda: Good point. I'd ignore object for now and handle only arrays.

src/Template/Layout/js/app/mixins/paginated-content.js
```diff
@@ -180,39 +180,46 @@ export const PaginatedContentMixin = {
...
+                    if (filterVal) {
+                        if (typeof filterVal === 'object') {
+                            return;
+                        }
+                        if (typeof filterVal === 'array') {
+                            if (filterVal.length > 0) {
+                                urlSearchParams.append(`filter[${filterKey}]`, filterVal);
+                            }
+                            return;
+                        }
```
le0m: this can be removed

didoda: done

PR #624, opened 2021-10-22
Add cakephp ide helper

didoda:
This adds the plugin https://github.com/dereuromark/cakephp-ide-helper

Available commands for `bin/cake`:
```
IdeHelper:
 - annotations
 - code_completion
 - illuminator
 - phpstorm
```

Bonus: annotate all files, with `bin/cake annotations all`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,10 +32,11 @@
     },
     "require-dev": {
         "cakephp/bake": "^1.11",
-        "cakephp/debug_kit": "^3.19.0",
         "cakephp/cakephp-codesniffer": "~3.3.0",
-        "psy/psysh": "@stable",
-        "phpunit/phpunit": "^6.5"
+        "cakephp/debug_kit": "^3.19.0",
+        "dereuromark/cakephp-ide-helper": "0.14.10",
+        "phpunit/phpunit": "^6.5",
+        "psy/psysh": "@stable"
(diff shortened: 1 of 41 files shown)


src/Application.php
```diff
@@ -56,6 +56,7 @@ public function bootstrap()
     protected function bootstrapCli()
     {
         parent::bootstrapCli();
+        $this->addPlugin('IdeHelper');
```
batopa: It's better to wrap `addPlugin()` in a `try catch` to avoid `MissingPluginException` in production env where `require-dev` are not installed.

batopa (changes requested):

composer.json
```diff
@@ -32,10 +32,11 @@
     },
     "require-dev": {
         "cakephp/bake": "^1.11",
-        "cakephp/debug_kit": "^3.19.0",
         "cakephp/cakephp-codesniffer": "~3.3.0",
-        "psy/psysh": "@stable",
-        "phpunit/phpunit": "^6.5"
+        "cakephp/debug_kit": "^3.19.0",
+        "dereuromark/cakephp-ide-helper": "0.14.10",
```
batopa: better  `"^0.14.10"`

batopa (approved):

Issue #630, opened 2021-10-26
Admin roles


didoda:
With #120 , the basic admin module (with applications, async jobs, config, endpoints, roles) is available, in readonly mode.

Enhance Admin Roles page, adding insert, modify and delete.

didoda:
Done with #630

PR #634, opened 2021-10-29
Roles save and remove

didoda:
This provides #630


src/Controller/Admin/RolesController.php
```diff
@@ -9,21 +12,59 @@
...
     protected $properties = ['name', 'description'];
+
+    /**
+     * Save roles
+     *
+     * @return \Cake\Http\Response|null
+     */
+    public function save(): ?Response
+    {
+        $this->request->allowMethod(['post']);
```
stefanorosanelli: This `save()` method and `remove()` could be placed in `AdministrationBaseController` they should be more or less the same in every context using `$this->resourceType` as parameter

stefanorosanelli (approved):

Issue #636, opened 2021-10-29
History info - application


didoda:
In modules view, History box, we have the list of history changes.
When we click on "Info" button, side panel shows history change details.
Among details, there should be the Application name.

In api response:
```
{
    "data": [
        {
            // ...
            "type": "history",
            "meta": {
                // ...
                "application_id": <id>,
                // ...
            }
        },
// ...
}
```

didoda:
done with #637

PR #637, opened 2021-10-29
Application name in history box

didoda:
This provides #636 

Requirement: https://github.com/bedita/bedita/pull/1836

### DIFF
--- a/src/Controller/Component/HistoryComponent.php
+++ b/src/Controller/Component/HistoryComponent.php
@@ -179,7 +179,8 @@ class HistoryComponent extends Component
                 $formatted[$field] = sprintf('%s: %s', __(Inflector::humanize($field)), $this->SchemaHelper->format($value, $fieldSchema));
             }
             $history['meta']['changed'] = $formatted;
-            $history['meta']['application_name'] = \App\Utility\Applications::getName($history['meta']['application_id']);
+            $applicationId = (string)Hash::get($history, 'meta.application_id');
+            $history['meta']['application_name'] = \App\Utility\Applications::getName($applicationId);
         }
         $response['data'] = $data;
     }
--- a/src/Controller/Component/HistoryComponent.php
+++ b/src/Controller/Component/HistoryComponent.php
@@ -1,6 +1,7 @@
 <?php
 namespace App\Controller\Component;
 
+use App\Utility\Applications;
 use App\View\Helper\SchemaHelper;
 use BEdita\WebTools\ApiClientProvider;
 use Cake\Controller\Component;
@@ -180,7 +181,7 @@ class HistoryComponent extends Component
             }
             $history['meta']['changed'] = $formatted;
             $applicationId = (string)Hash::get($history, 'meta.application_id');
-            $history['meta']['application_name'] = \App\Utility\Applications::getName($applicationId);
+            $history['meta']['application_name'] = Applications::getName($applicationId);
         }
         $response['data'] = $data;
     }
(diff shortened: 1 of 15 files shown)


src/Template/Layout/js/app/components/history/history.js
```diff
@@ -16,7 +16,10 @@ export default {
             <ul class="history-items">
                 <li class="history-item is-expanded py-05 has-border-gray-600" v-for="item in history[date]">
                     <div class="change-time"><: getFormattedTime(item.meta.created) :></div>
-                    <div class="is-flex">${t`by`} <a class="ml-05"><: getAuthorName(item.meta.user) :></a></div>
+                    <div class="is-flex">
+                        ${t`modified by`} <span class="has-font-weight-bold ml-05 mr-05"><: getAuthorName(item.meta.user) :></span>
```
stefanorosanelli: maybe a shorter version like `by {user} via {application}` ?

src/Utility/Applications.php
```diff
@@ -0,0 +1,60 @@
...
+     *
+     * @return array
+     */
+    public static function list(): array
+    {
+        try {
+            $applications = Cache::remember(
+                'applications',
+                function () {
+                    $response = (array)ApiClientProvider::getApiClient()->get('/admin/applications');
```
stefanorosanelli: if the user is session is not `admin` this API call and cache creation will fail

src/Controller/Component/HistoryComponent.php
```diff
@@ -179,6 +179,8 @@ private function formatResponseData(array &$response, array $schema): void
                 $formatted[$field] = sprintf('%s: %s', __(Inflector::humanize($field)), $this->SchemaHelper->format($value, $fieldSchema));
             }
             $history['meta']['changed'] = $formatted;
+            $applicationId = (string)Hash::get($history, 'meta.application_id');
+            $history['meta']['application_name'] = \App\Utility\Applications::getName($applicationId);
```
stefanorosanelli: import namespace using `use App\Utility\Applications`

src/Utility/Applications.php
didoda: does it? `/admin/applications` is not protected. The only protection happens in `ApplicationsController`

didoda: good point. how can we solve this? it seems to me that we need an endpoint with no restrictions that returns applications (id, name, description)

src/Template/Layout/js/app/components/history/history.js
didoda: `by` and `via` are "bad" strings for i18n, too generical and short [note that `{user}` and `{application}` are inside spans, so outside translation string].. we can create a `by %s via %s` translation and change the dom

src/Utility/Applications.php
stefanorosanelli: yes, something like that 
like `GET /config` for instance 

for now we could show this functionality to `admin` users only

stefanorosanelli (approved):
Ok for now - we could add `application` info also in `history-info` component (later also)

src/Utility/Applications.php
```diff
@@ -0,0 +1,60 @@
...
+     * Get applications, in the form [<id>: <name>].
+     * This uses cache 'applications'.
+     *
+     * @return array
+     */
+    public static function list(): array
+    {
+        try {
+            $applications = Cache::remember(
+                'applications',
```
stefanorosanelli: to support multi-project we should `use CacheTrait` and define cache key like:

```php
$this->cacheKey('applications')
```
see [SchemaComponent::getSchema](/bedita/manager/blob/master/src/Controller/Component/SchemaComponent.php#L60)

also a `protected static $applications` array could be used to avoid multiple cache hits

src/Utility/Applications.php
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * Get applications, in the form [<id>: <name>].
+     * This uses cache 'applications'.
+     *
+     * @return array
+     */
+    public static function list(): array
+    {
+        try {
```
stefanorosanelli, line 55: To avoid multiple cache hits simply add these lines first:

```php
if (!empty(static::$applications)) {
    return static::$applications
}
```

stefanorosanelli (approved):

PR #638, opened 2021-11-02
feat - Relations filters

DarioSacco:
This PR mainly adds filters for related objects panel.

When a relation has more than one right type, the object type select is showed. After selecting a type, the filter-box component shows the configured filters for that type.

Other refactoring and features in this PR:
- `urlUtils` helper, containing a method to build URL search params in a cleaner way. More methods can be added to this helper when needed
- new `capitalize` and `translate` Vue filters, that can help handle some strings in Vue context, also inside twig templates
- improved some old filter behavior (like the filter methods triggered on keyup of the search input)
- typos and other minor fixes

### DIFF
--- a/src/Template/Element/Panel/relations_add.twig
+++ b/src/Template/Element/Panel/relations_add.twig
@@ -10,7 +10,6 @@
         </header>
 
         <div class="create-new-object mt-1 mx-1" v-if="showCreateObjectForm">
-
             <div class="select">
                 <label for="related_object_type">{{ __('Type') }}</label>
                 <select id="related_object_type"
@@ -19,7 +18,7 @@
                     v-if="relationTypes"
                     v-model="object.type">
                     <option value="_choose">{{ __('Choose a type') }}</option>
-                    <option v-for="type in relationTypes.right"><: t(type) :></option>
+                    <option v-for="type in relationTypes.right"><: type | capitalize | translate :></option>
                 </select>
             </div>
 
@@ -40,7 +39,6 @@
 
                     <section class="fieldset mb-1">
                         <div class="container">
-
                             <fieldset id="{{ type }}-form-fields">
                                 {% set fields = config('Properties.' ~ type ~ '.fastCreate.all')|default(['status', 'title', 'description']) %}
                                 {% set required = config('Properties.' ~ type ~ '.fastCreate.required')|default(['status', 'title']) %}
@@ -53,7 +51,7 @@
                                         'key': type ~ '-' ~ field
                                     } %}
                                     {% if field == 'status' %}
-                                        {% set fieldOptions =  fieldOptions|merge({'v-model': 'object.attributes.status'}) %}
+                                        {% set fieldOptions = fieldOptions|merge({'v-model': 'object.attributes.status'}) %}
                                     {% endif %}
                                     {{ Property.control(field, '', fieldOptions, type)|raw }}
                                 {% endfor %}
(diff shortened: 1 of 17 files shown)


src/Template/Element/Panel/relations_add.twig
```diff
@@ -92,20 +90,30 @@
...
         <div class="px-1 my-1">
+            {% set list = {} %}
+            {% for type, filters in filtersByType %}
+                {% set options = [] %}
+                {% for f in filters %}
+                    {% set o = Schema.controlOptions(f, null, schema.properties[f]) %}
+                    {% set options = options|merge([o|merge({ 'name': f })]) %}
+                {% endfor %}
+                {% set list = list|merge({ (type): options }) %}
+            {% endfor %}
```
didoda, line 101: this part could be in an helper function (just to have more control over the logic, via unit tests)

src/Template/Layout/js/app/app.js
```diff
@@ -306,47 +306,11 @@ const _vueInstance = new Vue({
...
-                            const encoded = encodeURIComponent(query);
-                            fields.push(`${key}=${encoded}`);
-                        }
-                    }
-                });
-                url += fields.length ? queryId : '';
-                url += fields.join(separator);
-            }
+        buildUrlWithParams(params) {
+            let url = new URL(`${window.location.origin}${window.location.pathname}`);
```
didoda:
```suggestion
            const url = new URL(`${window.location.origin}${window.location.pathname}`);
```
minor, we can use `const` in here

src/Template/Layout/js/app/components/filter-box.js
```diff
@@ -139,7 +150,21 @@ export default {
...
+         * Fallback to the translation of "Items".
+         *
+         * @returns {String}
+         */
+        label() {
+            if (!this.objectsLabel) {
+                return t`Items`;
+            }
+
+            return t`${(this.objectsLabel.charAt(0) + this.objectsLabel.slice(1))}`;
```
didoda: we should avoid, if possible, expressions where the value to translate is a variable.
ttag extract cli won't find the string expressed by variable

```
t`items` => ok
t`this is ${var}` => ok
t`${var}` => ko
```

more example in https://ttag.js.org/docs/tag-gettext.html

src/Template/Layout/js/libs/filters.js
```diff
@@ -1,14 +1,38 @@
...
+});
+
+/**
+ * Translate a string.
+ *
+ * @param {String} str The string to translate
+ */
+Vue.filter('translate', function(str) {
+    return t([str]);
+});
```
didoda: ttag extract cli won't find strings to translate, referred by `'string' | translate`.
not sure this is a good idea

didoda (changes requested):
a lot of changes in `src/Template/Layout/js/app/components/filter-box.js`. hard to follow... and dangerous IMHO

src/Template/Layout/js/libs/filters.js
DarioSacco: `ttag extract` wouldn't find that string anyway, if it were in a variable (like often happens with the object type). This filter should be used only for dynamic strings, maybe I could make it clearer in the method jsdoc?

didoda: if you literally write in a template this:
```
{{ __('an example) }}
```
or in a js template:
```
<: t('an example') :>
```
it's ok. gettext shell will find `an example`.

if you write this:
```
<: 'an_example' | translate :>
```
it's not ok, because the component you added will always translate a string as a variable, even when it's not a dynamic variable.

that's why I don't think it's a good idea

DarioSacco: I know.
In the second case you should do this (if you just can't use `__('an_example')` for some reason)
```
<: t`an_example` :>
```

and use the filter only in cases like this
```
<: object.type | translate :>
```
I thought it would be useful because it can be chained with other filters, like `<: object.type | capitalize | translate :>`

src/Template/Element/Panel/relations_add.twig
DarioSacco: in the `SchemaHelper`?

src/Template/Layout/js/libs/filters.js
didoda: why not using directly t?
```
<: t(object.type | capitalize) :>
```

DarioSacco: I tried but filters don't work inside a function call 😞 

If we want to avoid introducing this filter, we can solve the specific situation (for object types) by adding lowercase translation keys for all object types...

didoda: I'd prefer `<: t(type) | capitalize :>` or `<: t(type) | humanize :>` and translating lowercase strings too, and removing `translate` filter.

PR #639, opened 2021-11-03
Categories tree in object view

didoda:
This provides https://github.com/bedita/manager/issues/640: categories groups, when categories are organized as tree (with parent_id not null in schema), by introducing `CategoriesHelper`. 

Categories without a parent and without children belongs to a special group "global", put at the beginning of categories box.

I.e.:

categories

Rif: https://chialab.atlassian.net/browse/BEM-63

### DIFF
--- a/src/Template/Element/Form/categories.scss
+++ b/src/Template/Element/Form/categories.scss
@@ -1,5 +1,12 @@
-.categories {
-    .select {
-        columns: 2 auto;
+.categories-container {
+    .categories {
+        margin-top: 15px;
+        .select {
+            columns: 2 auto;
+        }
+        h3 {
+            font-weight: bold;
+            text-transform: uppercase;
+        }
     }
 }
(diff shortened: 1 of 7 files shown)


src/Template/Element/Form/categories.scss
```diff
@@ -1,5 +1,12 @@
-.categories {
-    .select {
-        columns: 2 auto;
+.categories-container {
+    .categories {
+        margin-top: 15px;
```
DarioSacco:
```suggestion
        margin-top: $gutter;
```

src/View/Helper/CategoriesHelper.php
```diff
@@ -0,0 +1,169 @@
...
+            'hiddenField' => $hiddenField,
+        ];
+        $hiddenField = false;
+        if (empty($node['children'])) {
+            $controlOptions['options'][0] = ['value' => $node['name'], 'text' => $node['label']];
+
+            return $controlOptions;
+        }
+        $key = 0;
+        foreach ($node['children'] as $child) {
```
DarioSacco: why not using `$node['children'] as $index => $child` instead of the `$key` counter?

DarioSacco (approved):

PR #644, opened 2021-11-11
Fix show fastCreate form on type change

didoda:
This fix a problem in viewing fast create forms in side panel.

### DIFF
--- a/src/Template/Element/Panel/relations_add.twig
+++ b/src/Template/Element/Panel/relations_add.twig
@@ -18,7 +18,7 @@
                     v-if="relationTypes"
                     v-model="object.type">
                     <option value="_choose">{{ __('Choose a type') }}</option>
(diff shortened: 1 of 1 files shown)


src/Template/Element/Panel/relations_add.twig
```diff
@@ -18,7 +18,7 @@
                     v-if="relationTypes"
                     v-model="object.type">
                     <option value="_choose">{{ __('Choose a type') }}</option>
-                    <option v-for="type in relationTypes.right"><: t(capitalize(type)) :></option>
+                    <option v-for="type in relationTypes.right" v-bind:value="type"><: t(capitalize(type)) :></option>
```
DarioSacco:
```suggestion
                    <option v-for="type in relationTypes.right" :value="type"><: t(capitalize(type)) :></option>
```

`v-bind` is not necessary. Using `:` before the attribute is enough

DarioSacco (approved):

PR #651, opened 2021-11-22
Async jobs, enable delete

### LINKED ISSUE #628: Admin async jobs
With #120 , the basic admin module (with applications, async jobs, config, endpoints, roles) is available, in readonly mode.

Provide async jobs delete in Admin Async Jobs page.

didoda:
This resolves #628  providing "delete" for async jobs.

### DIFF
--- a/src/Controller/Admin/AdministrationBaseController.php
+++ b/src/Controller/Admin/AdministrationBaseController.php
@@ -35,6 +35,13 @@ abstract class AdministrationBaseController extends AppController
     protected $readonly = true;
 
     /**
+     * Deleteonly flag view.
+     *
+     * @var bool
+     */
+    protected $deleteonly = false;
+
+    /**
      * Properties to show in index columns
      *
      * @var array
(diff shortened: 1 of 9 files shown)


src/Controller/Admin/AdministrationBaseController.php
```diff
@@ -34,6 +34,13 @@ abstract class AdministrationBaseController extends AppController
      */
     protected $readonly = true;
 
+    /**
+     * Deleteonly flag view.
+     *
+     * @var bool
+     */
+    protected $deleteonly = false;
```
stefanorosanelli, line 42: Maybe `$allowDelete` or similar? And `'allowDelete'` or similar as string/var below

(sounds like "only delete operation is possible")

didoda: the meaning is exactly: "allow only delete for this page"

stefanorosanelli (approved):

PR #652, opened 2021-11-22
Edit module categories

edoardocavazza:
Enable per module categories list (and editing).

### DIFF
--- a/src/Controller/Model/CategoriesController.php
+++ b/src/Controller/Model/CategoriesController.php
@@ -45,14 +45,12 @@ class CategoriesController extends ModelBaseController
     {
         $features = $this->Schema->objectTypesFeatures();
         $categorized = (array)Hash::get($features, 'categorized');
-        $schema = $this->Schema->getSchema();
-        $schema['properties']['type']['enum'] = $categorized;
         $filter = (array)$this->request->getQuery('filter');
         if (!empty($filter['type'])) {
             $categorized = [$filter['type']];
         }
-        $categorized = array_combine($categorized, $categorized);
-        $this->set(compact('categorized', 'schema'));
+
+        $this->set('object_types', array_values($categorized));
 
         return parent::beforeRender($event);
     }
--- a/src/Controller/Model/CategoriesController.php
+++ b/src/Controller/Model/CategoriesController.php
@@ -12,13 +12,10 @@
  */
 namespace App\Controller\Model;
 
-use BEdita\SDK\BEditaClientException;
 use Cake\Http\Response;
-use Cake\Utility\Hash;
 
 /**
  * Categories Model Controller: list, add, edit, remove categories
- *
  *
  * @property \App\Controller\Component\CategoriesComponent $Categories
  * @property \App\Controller\Component\PropertiesComponent $Properties
--- a/src/Controller/Model/CategoriesController.php
+++ b/src/Controller/Model/CategoriesController.php
@@ -14,6 +14,7 @@ namespace App\Controller\Model;
 
 use BEdita\SDK\BEditaClientException;
 use Cake\Http\Response;
+use Cake\Utility\Hash;
 
 /**
  * Categories Model Controller: list, add, edit, remove categories
@@ -44,7 +45,7 @@ class CategoriesController extends ModelBaseController
     {
         $this->request->allowMethod(['get']);
         $query = $this->request->getQueryParams() + [
-            'page_size' => 500,
+            'page_size' => 100,
         ];
 
         try {
@@ -53,13 +54,10 @@ class CategoriesController extends ModelBaseController
             $this->log($e, 'error');
             $this->Flash->error($e->getMessage(), ['params' => $e]);
 
-            return $this->redirect(['_name' => 'dashboard']);
+            return $this->redirect($this->referer());
         }
 
-        $resources = [];
-        foreach ((array)$response['data'] as $category) {
-            $resources[$category['id']] = $category;
-        }
+        $resources = (array)Hash::combine((array)$response['data'], '{n}.id', '{n}');
 
         $grouped = [
             '_' => [],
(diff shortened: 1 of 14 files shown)


src/Controller/Model/CategoriesController.php
```diff
@@ -39,21 +38,50 @@ class CategoriesController extends ModelBaseController
...
-        $filter = (array)$this->request->getQuery('filter');
-        if (!empty($filter['type'])) {
-            $categorized = [$filter['type']];
+        $this->request->allowMethod(['get']);
+        $query = $this->request->getQueryParams() + [
+            'page_size' => 500,
+        ];
+
+        try {
+            $response = $this->apiClient->get('/model/categories', $query);
```
didoda: 500 categories could be all, or not. 
you could do something as follows:
```
$query = $this->request->getQueryParams() + [
    'page_size' => 500,
];
try {
    $response = $this->apiClient->get('/model/categories', $query);
    $data = (array)$response['data'];
    $pageCount = $response['meta']['pagination']['page_count'];
    if ($pageCount > 1) {
        for ($i = 2; $i < $pageCount; $i++) {
            $query['page'] = $i;
            $response = $this->apiClient->get('/model/categories', $query);
            $data = array_merge($data, (array)$response['data']);
        }
    }
}
// then use $data instead of $response['data']
```

src/Controller/Model/CategoriesController.php
```diff
@@ -39,21 +38,50 @@ class CategoriesController extends ModelBaseController
...
+        } catch (BEditaClientException $e) {
+            $this->log($e, 'error');
+            $this->Flash->error($e->getMessage(), ['params' => $e]);
+
+            return $this->redirect(['_name' => 'dashboard']);
```
didoda: I'd redirect to referer, instead.
```
return $this->redirect($this->referer());
```

src/Controller/Model/CategoriesController.php
```diff
@@ -39,21 +38,50 @@ class CategoriesController extends ModelBaseController
...
+        }
+
+        $resources = [];
+        foreach ((array)$response['data'] as $category) {
+            $resources[$category['id']] = $category;
         }
```
didoda: you can do it with
```
$resources = (array)Hash::combine((array)$response['data'], '{n}.id', '{n}');
```
if you use `$data` as suggested above, then:
```
$resources = (array)Hash::combine($data, '{n}.id', '{n}');
```

src/Controller/Model/CategoriesController.php
```diff
@@ -39,21 +38,50 @@ class CategoriesController extends ModelBaseController
...
+        foreach ($resources as $category) {
+            if (empty($category['attributes']['parent_id'])) {
+                $grouped['_'][] = $category['id'];
+            } else {
+                $grouped[$category['attributes']['parent_id']][] = $category['id'];
+            }
+        }
+
+        $object_types = $this->Schema->objectTypesFeatures()['categorized'];
+        $object_types = array_combine($object_types, $object_types);
```
didoda: minor, use `$objectTypes`

src/Controller/ModulesController.php
```diff
@@ -523,4 +523,117 @@ public function getSchemaForIndex($objectType): array
...
+        $this->request->allowMethod(['get']);
+        $query = $this->request->getQueryParams() + [
+            'filter' => [
+                'type' => $this->objectType,
+            ],
+            'page_size' => 500,
+        ];
+
+        try {
+            $response = $this->apiClient->get('/model/categories', $query);
```
didoda: as above. there could be more than 500 categories...

src/Controller/ModulesController.php
```diff
@@ -523,4 +523,117 @@ public function getSchemaForIndex($objectType): array
...
+        } catch (BEditaClientException $e) {
+            $this->log($e, 'error');
+            $this->Flash->error($e->getMessage(), ['params' => $e]);
+
+            return $this->redirect(['_name' => 'dashboard']);
```
didoda: maybe better redirect to referer

src/Controller/ModulesController.php
```diff
@@ -523,4 +523,117 @@ public function getSchemaForIndex($objectType): array
...
+        }
+
+        $resources = [];
+        foreach ((array)$response['data'] as $category) {
+            $resources[$category['id']] = $category;
+        }
```
didoda: as above, you can use `Hash::combine`

src/Controller/ModulesController.php
```diff
@@ -523,4 +523,117 @@ public function getSchemaForIndex($objectType): array
...
+        $this->set(compact('resources', 'grouped'));
+        $this->set('meta', (array)$response['meta']);
+        $this->set('links', (array)$response['links']);
+        $this->set('schema', $this->Schema->getSchema());
+        $this->set('properties', $this->Properties->indexList('categories'));
+        $this->set('filter', $this->Properties->filterList('categories'));
+        $this->set('object_types', [$this->objectType]);
+
+        return null;
+    }
```
didoda, line 553: the logic in this function is very similar to `index` (except for `$query`). you could put it in a function inside a component or in a model

didoda (changes requested):

src/Controller/Model/CategoriesController.php
edoardocavazza: Categories index is paginated, we don't need to load all categories at once. Anyway, 500 is not a valid value for BEdita because the maximum is 100 I think. I'll fix it.

src/Controller/ModulesController.php
edoardocavazza: Yes, I was not sure how to abstract this part. The component solution should fit I think

didoda (approved):

didoda (changes requested):
It seems ok.

Maybe better add some tests for:

 - CategoriesComponent::index
 - CategoriesController::initialize
 - CategoriesController::index
 - ModulesController::listCategories
 - ModulesController::saveCategory
 - ModulesController::removeCategory

Minor: methods about Categories in ModulesController could be moved to Categories Component (ModulesController is already "crowded" :) ).

didoda (approved):

PR #654, opened 2021-11-25
Enhance history

### LINKED ISSUE #460: Enhance history info panel
History info panel shows changed data, property name and correspondant value.
Some enhancement could be done:

 - better style: values are rendered as html, and sometimes it results ugly to see (unexpected new lines, etc.)
 - ad hoc properties: "date ranges" should be seen in a custom way

didoda:
This resolves #460 providing a better view for history info panel.

Custom view for:

 - date ranges
 - categories
 - coordinates

Changes not shown with fields:

 - model-type
 - _csrfToken

### DIFF
--- a/src/Template/Element/Form/form.scss
+++ b/src/Template/Element/Form/form.scss
@@ -13,7 +13,7 @@
 
         .history-field {
             border: dashed gray 1px;
-            padding: 0.5rem 0.5rem 0.5rem 0.5rem;
+            padding: $gutter * .5;
         }
 
         .history-field > label {
--- a/src/Template/Element/Form/form.scss
+++ b/src/Template/Element/Form/form.scss
@@ -11,6 +11,15 @@
             }
         }
 
+        .history-field {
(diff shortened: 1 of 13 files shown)


src/Template/Element/Form/form.scss
```diff
@@ -11,6 +11,15 @@
             }
         }
 
+        .history-field {
+            border: dashed gray 1px;
+            padding: 0.5rem 0.5rem 0.5rem 0.5rem;
```
DarioSacco:
```suggestion
            padding: $gutter * .5;
```

src/Template/Element/Form/form.scss
```diff
@@ -11,6 +11,15 @@
             }
         }
 
+        .history-field {
+            border: dashed gray 1px;
```
DarioSacco: why not use a gray sass variable, like `$gray-600`, from `_variables.scss`?

DarioSacco (approved):

PR #656, opened 2021-12-01
Properties init and cache

### LINKED ISSUE #655: Properties.<module> wrongly overrides defaults
When `Properties.<module>` has its own keys, we expect that the contents are merged with defaults.

For example, when config is as follows:
```php
'Properties' => [
        'users' => [
            'fastCreate' => [
                'required' => ['status', 'username'],
                'all' => ['status', 'username', 'email'],
            ],
    ],
],
```
At runtime (`PropertiesComponent::initialize()`) we expect a merged with defaults "properties" config, like:
```php
'Properties' => [
        'users' => [
            '_keep' => [
                // defaults ...
            ],
            'fastCreate' => [
                'required' => ['status', 'username'],
                'all' => ['status', 'username', 'email'],
            ],
            'index' => [
                // defaults ...
            ],
            'view' => [
                // defaults ...
            ],
    ],
],
```

didoda:
This fixes #655 

Bonus: caching the properties, so that the properties init is done just once, and not every time the component is used.

### DIFF
--- a/src/Controller/Component/PropertiesComponent.php
+++ b/src/Controller/Component/PropertiesComponent.php
@@ -13,6 +13,8 @@
 
 namespace App\Controller\Component;
 
+use App\Utility\CacheTools;
+use Cake\Cache\Cache;
 use Cake\Controller\Component;
 use Cake\Core\Configure;
 use Cake\Utility\Hash;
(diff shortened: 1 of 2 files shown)


src/Controller/Component/PropertiesComponent.php
```diff
@@ -84,12 +86,45 @@ class PropertiesComponent extends Component
...
     }
 
+    /**
+     * Init properties
+     *
+     * @return void
+     */
+    protected function init(): void
+    {
+        $cacheKey = CacheTools::cacheKey('properties');
```
stefanorosanelli, line 101: Overall ok, `Cache` usage seems like overkill in order to avoid these array manipulations

stefanorosanelli (approved):

src/Controller/Component/PropertiesComponent.php
didoda: well... it's called 4 times on every content "view". not a lot, but I'd say better avoiding any unrequired repetition

nicolocarpignoli (approved):

PR #665, opened 2022-01-03
Update dependencies: i18n 1.7, web-tools 1.8 and cakephp 3.10.*

didoda:
This updates dependencies:

 - "bedita/i18n": "^1.7"
 - "bedita/web-tools": "^1.8"
 - "cakephp/bake": "^1.12"

New dependency:

 - "cakephp/cakephp": "^3.10"
 - "phpstan/phpstan": "^1.3" (dev dependency)

Bonus:

 - update workflow javascript to launch checks on `.github/workflows/javascript.yml` and `package.json` changes too
 - update workflow php to launch checks on `.github/workflows/php.yml` and `composer.json` changes too


composer.json
```diff
@@ -22,19 +22,21 @@
     ],
     "require": {
         "php": ">=7.1",
-        "bedita/i18n": "^1.6",
-        "bedita/web-tools": "^1.7.1",
+        "bedita/i18n": "^1.7",
+        "bedita/web-tools": "^1.8",
+        "cakephp/cakephp": "^3.10",
```
stefanorosanelli, line 27: since in `bedita/web-tools` 1.8 we already require `"cakephp/cakephp": "3.10.*"` I don't think the `cakephp/cakephp` dependency is necessary

didoda: not necessary, but a good way to have a version specified in main app (@batopa suggested an explicit dependency here)

batopa: When an item of a dependency is used in the main code base is a best practice to explicit the dependency in `composer.json`. This makes more clear what are the dependencies of the app and give more control on specific version used.

stefanorosanelli (approved):

PR #667, opened 2022-01-07
composer.lock versioning [standby - could be problematic on plugins]

didoda:
This puts composer.lock under version control.

To avoid workflow errors, we assume to use php 7.4 only

Bonus: more coverage on 2 classes


didoda:
This could break compatibility with plugins. Closing for now

PR #679, opened 2022-01-18
modules order in dashboard

### LINKED ISSUE #678: Modules list order in Dashboard and other pages
Modules list order should follow config `Modules` order in Dashboard and other pages (top menu).

didoda:
This resolves #678

### DIFF
--- a/src/Template/Pages/Dashboard/index.twig
+++ b/src/Template/Pages/Dashboard/index.twig
@@ -7,46 +7,30 @@
             <div class="dashboard-items">
             {% for name, module in modules if not in_array(name, ['trash', 'users']) %}
 
+                {# Special cases - set defaults on objects, publications, folders #}
                 {% if name == 'objects' %}
-
-                    {{ Html.link(__('All objects'),
-                    { '_name': 'modules:list', 'object_type': 'objects' },
-                    { 'title': __('All objects'), 'class': 'dashboard-item has-background-module-objects icon-th-large-1', } )|raw }}
-
+                    {% set module = {'label': __('All objects')}|merge(module) %}
                 {% elseif name == 'publications' %}
-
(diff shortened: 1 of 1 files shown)


src/Template/Pages/Dashboard/index.twig
```diff
@@ -5,57 +5,51 @@
...
-                {% if module.route %}
-                    {% set url = Url.build(module.route) %}
-                {% else %}
-                    {% set url = Url.build({ '_name': 'modules:list', 'object_type': name, 'plugin': null }) %}
-                {% endif %}
-                <a href="{{ url }}" class="{{ class }}">
-                    <span>{{ title }}</span>
-                </a>
-            {% endfor %}
+                {% if name == 'objects' %}
```
stefanorosanelli, line 11: I suggest this change, since in every module `label` and `class` can be defined in configuration, we are allowed to override them in every module.

```twig 
                {# Special cases - set defaults on objects, publications, folders #}

                {% if name == 'objects' %}
                    {% set module = {'label': 'All objects'}|merge(module) %}
                {% elseif name == 'publications' %}
                    {% set module = {'class': 'icon-globe-1'}|merge(module) %}
                {% elseif name == 'folders' %}
                    {% set module = {'class': 'icon-folder'}|merge(module) %}
                {% endif %}

                {% set title = Layout.typeLabel(module.label|default(name)|humanize) %}

                {% set class = 'dashboard-item has-background-module-%s %s'|format(name, module.class|default('')) %}
                {% if module.hints.multiple_types and not module.class %}
                    {% set class = class ~ ' icon-th-large-1' %}
                {% endif %}

```
then unchanged from line 
same logic could be applied in modules below, we could also use an element for this part

stefanorosanelli (approved):

PR #683, opened 2022-01-27
Show categories buttons only when in schema

didoda:
This fixes an unexpected behavior, that make categories buttons always visible in object index and view.
Categories buttons should be visible only when schema contains 'categories'.

### DIFF
--- a/src/Template/Pages/Model/Categories/index.twig
+++ b/src/Template/Pages/Model/Categories/index.twig
@@ -10,7 +10,7 @@
 {% element 'Modules/index_header' %}
 
 <div class="module-index">
-    {% if schema.categories %}
+    {% if schema.properties.categories %}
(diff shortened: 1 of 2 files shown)


src/Template/Pages/Model/Categories/index.twig
```diff
@@ -10,10 +10,12 @@
 {% element 'Modules/index_header' %}
 
 <div class="module-index">
+    {% if schema.properties.categories %}
```
stefanorosanelli: With this `{⅜ if .. %}` I don't see any category... is it necessary?

stefanorosanelli (approved):

src/Template/Pages/Model/Categories/index.twig
didoda: restored as before, this part

PR #691, opened 2022-02-04
refactor: replace AuthComponent with Authentication plugin

le0m:
This PR aims to replace `AuthComponent` (which has been [deprecated in CakePHP 4](https://github.com/cakephp/cakephp/blob/4.x/src/Controller/Component/AuthComponent.php#L43)) with [Authentication plugin](https://github.com/cakephp/authentication) (which is suggested and will supersede it).

This PR uses v1.x of the Authentication plugin because it's the version compatible with CakePHP 3. We will switch to v2.x when we'll make the switch to CakePHP 4.


didoda (approved):

src/Application.php
```diff
@@ -155,4 +166,36 @@ public static function loadProjectConfig(?string $project, string $projectsPath)
             Configure::load($project, 'projects');
         }
     }
+
+    /**
+     * @inheritDoc
+     */
+    public function getAuthenticationService(ServerRequestInterface $request, ResponseInterface $response)
```
didoda:
```suggestion
    public function getAuthenticationService(ServerRequestInterface $request, ResponseInterface $response): AuthenticationService
```
Add return type to function

src/Controller/AppController.php
```diff
@@ -150,9 +149,12 @@ protected function loginRedirectRoute(): array
...
     {
-        $timezone = $this->Auth->user('timezone');
-        if ($timezone) {
-            Configure::write('I18n.timezone', $timezone);
+        $identity = $this->Authentication->getIdentity();
+        if ($identity) {
+            $timezone = $identity->get('timezone');
+            if ($timezone) {
+                Configure::write('I18n.timezone', $timezone);
+            }
```
didoda: minor. we could reduce method complexity using "return first" pattern.
i.e.:
```
$identity = $this->Authentication->getIdentity();
if (!$identity) {
   return;
}
$timezone = $identity->get('timezone');
if (!$timezone) {
    return;
}
Configure::write('I18n.timezone', $timezone);
```

src/Controller/AppController.php
```diff
@@ -163,13 +165,13 @@ protected function setupOutputTimezone(): void
...
             $tokens = $this->apiClient->getTokens();
-            if ($tokens && $user['tokens'] !== $tokens) {
-                // Update tokens in session.
-                $user['tokens'] = $tokens;
-                $this->Auth->setUser($user);
+            if ($tokens && $user->get('tokens') !== $tokens) {
+                $data = compact('tokens') + (array)$user->getOriginalData();
+                $user = new Identity($data);
+                $this->Authentication->setIdentity($user);
             }
```
didoda, line 175: as in previous comment... "return first" approach could reduce complexity

didoda (approved):

PR #703, opened 2022-03-01
Tag management on object detail

nicolocarpignoli:
Introducing `tag-picker` on object details in order to:

- show current tags added to the object
- remove tags
- add available tags, searching with autocomplete function
- add a new tag to the object (and to the global tags list).

### DIFF
--- a/src/Template/Element/Form/tags.twig
+++ b/src/Template/Element/Form/tags.twig
@@ -1,0 +1,21 @@
+
+{# {% if schema.tags is defined %} #}
+{# TODO understand why this does not work #}
+    <property-view inline-template :tab-open="tabsOpen" tab-name="tags">
+
+        <section class="fieldset">
+            <header @click.prevent="toggleVisibility()"
+                class="tab unselectable"
+                :class="isOpen? 'open has-border-module-{{ currentModule.name }}' : ''">
+                <h2>{{ __('Tags') }}</h2>
+            </header>
+
+            <div v-show="isOpen" class="tab-container">
+                <div class="tags-container">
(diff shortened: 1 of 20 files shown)


didoda (approved):

src/Template/Layout/js/app/components/tag-picker/tag-picker.js
```diff
@@ -0,0 +1,114 @@
...
+
+    mounted() {
+        this.selectedTags = this.initialTags?.map((tag) => ({ id: tag.name, label: tag.label }));
+        this.parseBeforeSave();
+    },
+
+    methods: {
+        parseBeforeSave() {
+            this.modifiedTags = JSON.stringify(this.selectedTags.map((tag) => ({ name : tag.id, label: tag.label })));
+        },
```
DarioSacco, line 69: I think you can use a computed property instead of this method and the `modifiedTags` data prop. some other adjustment is required to trigger the lifecycle on `selectedTags` changes (replacing `splice` instead of `push` should work)

src/Template/Element/Form/tags.twig
```diff
@@ -0,0 +1,16 @@
+{% if 'Tags' in schema.associations %}
+    <property-view inline-template :tab-open="tabsOpen" tab-name="tags">
+        <section class="fieldset">
+            <header @click.prevent="toggleVisibility()"
+                class="tab unselectable"
+                :class="isOpen? 'open has-border-module-{{ currentModule.name }}' : ''">
```
DarioSacco, line 6:
```suggestion
                :class="isOpen ? 'open has-border-module-{{ currentModule.name }}' : ''">
```

DarioSacco (changes requested):

PR #719, opened 2022-03-15
Refactor, cake4 phpcs compatibility

didoda:
This updates phpcs rules and applies a refactor.


.github/workflows/php.yml
```diff
@@ -50,11 +50,21 @@ jobs:
       - name: 'Install dependencies with Composer'
         run: 'composer install --prefer-dist --no-interaction'
 
+      - name: 'Setup PHP CodeSniffer - installed_paths'
```
stefanorosanelli: are these necessary? these settings are already in `phpcs.yml` config file

stefanorosanelli (approved):

.github/workflows/php.yml
didoda: Apparently `phpcs.yml` is ignored... When setup cs is not performed, phpcs workflow step fails with several errors

didoda: `phpcs.yml` should be `phpcs.xml`. my bad... fix soon

PR #741, opened 2022-03-28
Update phpstan/phpstan to version 1.5 + fix php workflow

didoda:
This updates `php` workflow, so that `phpstan` dependency version is used (instead of downloading it with composer, during CI).

This activity got the idea from https://github.com/bedita/bedita/pull/1889

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -1,0 +1,7 @@
+parameters:
+  bootstrapFiles:
+    - tests/bootstrap.php
+  paths:
+    - src
+    - tests
+  level: 0
--- a/.github/workflows/php.yml
+++ b/.github/workflows/php.yml
@@ -94,7 +94,7 @@ jobs:
 
       - name: 'Run PHP STAN'
         run: |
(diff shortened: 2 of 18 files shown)


stefanorosanelli (approved):

fquffio (approved):

phpstan.neon.dist
```diff
@@ -0,0 +1,7 @@
+parameters:
+  bootstrapFiles:
+    - tests/bootstrap.php
+  paths:
+    - src
+    - tests
+  level: 0
```
fquffio, line 7: This is the default and most basic level. I see that this level already highlighted a few issues in test cases.

I'd be curious to see if we can raise the bar a little bit. Levels are 0-9, but I think attempting a level beyond 6 on an existing and complex project could turn out to be a nightmare.

Would you be willing to try and experiment a little with levels 1, 2, 3, …? I don't expect a level increase within this PR, of course, but just to have an idea of the amount of work needed…

didoda: I would give it a try (a deeper level change) in a separate PR

PR #742, opened 2022-03-29
phpstan level 2

didoda:
After https://github.com/bedita/manager/pull/741 phpstan level is set at 0, in `phpstan.neon.dist`.

This provides a refactor, updating level to an higher number.

### DIFF
--- a/src/Controller/Component/ModulesComponent.php
+++ b/src/Controller/Component/ModulesComponent.php
@@ -59,7 +59,7 @@ class ModulesComponent extends Component
      *
      * @var array
      */
-    public $modules = [];
+    protected $modules = [];
 
     /**
      * Read modules and project info from `/home' endpoint.
--- a/src/Controller/Component/ModulesComponent.php
+++ b/src/Controller/Component/ModulesComponent.php
@@ -68,9 +68,9 @@ class ModulesComponent extends Component
      */
     public function startup(): void
     {
-        /** @var \Authentication\Identity $user */
+        /** @var \Authentication\Identity|null $user */
         $user = $this->Authentication->getIdentity();
-        if (!$user->get('id')) {
+        if (empty($user) || !$user->get('id')) {
             $this->getController()->set(['modules' => [], 'project' => []]);
 
             return;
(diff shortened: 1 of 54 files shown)


src/Controller/Component/ModulesComponent.php
```diff
@@ -68,8 +68,9 @@ class ModulesComponent extends Component
      */
     public function startup(): void
     {
+        /** @var \Authentication\Identity $user */
```
fquffio: I see a lot of these annotations in this PR. What's the type that is automatically inferred by phpstan?

I think that in some cases the correct type to document would be `\Authentication\Identity|null`, which explains why there used to be a few `empty($user)` checks…

src/Controller/ErrorController.php
```diff
@@ -61,15 +61,4 @@ public function beforeRender(EventInterface $event): ?Response
...
     }
-
-    /**
-     * afterFilter callback.
-     *
-     * @param \Cake\Event\EventInterface $event Event.
-     * @return \Cake\Http\Response|null|void
-     * @codeCoverageIgnore
-     */
-    public function afterFilter(EventInterface $event): void
```
fquffio, line 72: This method was overridden here to ensure nothing happens in `afterFilter` that may interfere with error rendering…

tests/TestCase/Controller/Component/ModulesComponentTest.php
```diff
@@ -91,7 +91,8 @@ public function objectTypes(?bool $abstract = null): array
                 return ['mices', 'elefants', 'cats', 'dogs'];
             }
         };
-        $controller->Authentication = $this->Authentication;
+        $controller->loadComponent('Authentication');
+        //$controller->Authentication = $this->Authentication;
```
fquffio:
```suggestion
```

tests/TestCase/Controller/Component/ObjectsEditorsComponentTest.php
```diff
@@ -134,7 +134,8 @@ public function testUpdate(): void
 
         // id not found in objectsEditors
         $user = ['id' => 123, 'attributes' => ['name' => 'Gustavo', 'surname' => 'Support']];
-        $this->ObjectsEditors->getController()->Authentication->setIdentity(new Identity($user));
+        //$controller->Authentication = $this->Authentication;
```
fquffio:
```suggestion
```

src/Controller/Component/ModulesComponent.php
didoda: The inferred type was the interface. Good point on using `\Authentication\Identity|null` instead of `Authentication\Identity`

PR #746, opened 2022-03-31
Refactor webpack - alternate paths for locales, templates and javascript root

didoda:
This provides a way to check alternate multiple paths for locales, templates and javascript root, when building manager and plugins.

Default paths:

```
jsRoot: 'src/Template/Layout/js',
templateRoot: 'src/Template',
localeDir: 'src/Locale',
```

Alternate paths:

```
alternateJsRoots: ['templates/Layout/js', 'templates/layout/js', 'resources/js'],
alternateTemplateRoots: ['templates/layout', 'templates'],
alternateLocaleDirs: ['locales'],
```

For each type of config (locales, templates and javascript root), during build it checks if default path exists; if it doesn't, use alternate path instead, if find one available (checking folder exists).

This fixes a compatibility issue on using a plugin that uses `templates` path. This is also a step to a future "move" of files, more cake4 compliant.

### DIFF
--- a/webpack.config.environment.js
+++ b/webpack.config.environment.js
@@ -12,7 +12,7 @@ const { readdirSync, statSync, existsSync } = require('fs')
 const { join } = require('path')
 
 const readDirs = p => readdirSync(p).filter(f => statSync(join(p, f)).isDirectory());
-const dirExists = p => existsSync(p);
+const fileExists = p => existsSync(p);
 
 // Environment: default development
 // from Node env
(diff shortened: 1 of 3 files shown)


webpack.config.environment.js
```diff
@@ -93,6 +94,11 @@ const BUNDLE = {
     localeDir: 'src/Locale',
     beditaPluginsRoot: 'plugins',
 
+    // alternate folders, cake4 pattern
+    alternateJsRoot: 'templates/Layout/js',
```
stefanorosanelli: Template layout folder in Cake 4 is `templates/layout` 

Javascript files could be set somewhere else, in many projects we use `resources/js` for instance as js root and use `templates` for `.twig` files only

didoda: what would you suggest to do? having a set of alternate paths, is a possibility? i.e. `alternateJsRoots: ['templates/Layout/js', 'templates/layout/js', 'resources/js'],`

stefanorosanelli: If this is an alternate root using the new Cake 4 folder layout I would probably use only `'resources/js'`

didoda: good point, but not working with existing plugins that use `templates/Layout/js`. The main idea of this PR was to maintain retrocompatibility (fix bug for those plugins introduced by manager passage to cake4) plus moving forward a step towards cake 4 folder layout compatibility

stefanorosanelli (approved):

PR #751, opened 2022-04-05
fix: use `?` in route array

stefanorosanelli:
This PR fixes a problem when a user logout happens caused by session expiration. 
The new `/login` URL must have a `redirect` query string used to return to the same location after a successful login. 

In Cake 4 `Router::url(['_name' => 'login', 'redirect' => '/documents'])` is not returning the expected URL `/login?redirect=/documents` like in Cake 3 and apparently like the official Cake 4 documentation states https://book.cakephp.org/4/en/development/routing.html#using-named-routes

### DIFF
--- a/tests/TestCase/Authentication/Identifier/Resolver/ApiResolverTest.php
+++ b/tests/TestCase/Authentication/Identifier/Resolver/ApiResolverTest.php
@@ -123,6 +123,7 @@ class ApiResolverTest extends TestCase
         $token = $identity['tokens']['renew'];
         $identity = $resolver->find(compact('token'));
 
+        unset($expected['tokens'], $identity['tokens']);
         foreach ($expected as $key => $val) {
             $this->assertEquals($val, $identity[$key]);
         }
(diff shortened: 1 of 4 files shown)


didoda (approved):

tests/TestCase/Authentication/Identifier/Resolver/ApiResolverTest.php
```diff
@@ -123,6 +123,7 @@ public function testFind(array $credentials, ?array $expected): void
         $token = $identity['tokens']['renew'];
         $identity = $resolver->find(compact('token'));
 
+        unset($expected['tokens'], $identity['tokens']);
```
fquffio: Rather than unsetting these values here, I'd remove the pieces of code that populate these values in the first place:

- lines 107-110 (`if isset($expected['tokens']) { ... }`)
- line 86 (`'tokens' => [],`)

stefanorosanelli: Right, done

didoda (approved):

PR #755, opened 2022-04-11
fix: Check if response exists before calling hash function

giacomor91:
Check if api client response exists before returning thumbs data. If response is null, returns empty array and prevent error that won't let any children folder element to show up.

### DIFF
--- a/src/Controller/Component/ThumbsComponent.php
+++ b/src/Controller/Component/ThumbsComponent.php
@@ -91,6 +91,9 @@ class ThumbsComponent extends Component
             ]));
             $apiClient = ApiClientProvider::getApiClient();
             $res = $apiClient->get($url, $query);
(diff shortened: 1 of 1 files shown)


src/Controller/Component/ThumbsComponent.php
```diff
@@ -91,6 +91,9 @@ protected function getThumbs(array $ids): ?array
             ]));
             $apiClient = ApiClientProvider::getApiClient();
             $res = $apiClient->get($url, $query);
+            if ($res == null) {
+                return [];
+            }
```
didoda, line 94: `$res = (array)$apiClient->get($url, $query);` should do the job (and we avoid an extra path, caused by `if`)

didoda (approved):

PR #759, opened 2022-04-20
Custom labels

didoda:
This provides a way to customize labels.

You can customize the labels for a control setting a configuration `Properties.$objectType.labels` as follows.
```
'Properties' => [
    // ...
    'documents' => [
        // ...
        'labels' => [
            'fields' => [
                'description' => 'D E S C R I P T I O N',
            ],
            'options' => [
                'related_stuff' => [
                    81231 => 'Dummy article',
                    12156 => 'Interesting stuff',
                    15975 => 'Whatever',
                ],
            ],
        ],
    ],
],

// resulting description field
<label for="description">D E S C R I P T I O N</label>
// ...

// resulting combo will be something like
<select name="related_stuff">
    <option value="281231">Dummy article</option>
    <option value="12156">Interesting stuff</option>
    <option value="15975">Whatever</option>
</select>
```

### DIFF
--- a/config/app.default.php
+++ b/config/app.default.php
@@ -515,8 +515,8 @@ return [
      *  - 'bulk' bulk actions list
      *  - 'fastCreate' fields for fast creation forms, by type
      *  - 'labels' to customize labels of properties names and their options, it can have two keys `fields` and `options`
-         + 'fields' contains an associative array with property names as keys and labels as values 
-         + 'options' contains an associative array with property names as keys and an array containing option items and the custom labels to use
+     *   + 'fields' contains an associative array with property names as keys and labels as values
+     *   + 'options' contains an associative array with property names as keys and an array containing option items and the custom labels to use
      * A special custom element 'Form/empty' can be used to hide a property group or relation via `_element`
      */
     // 'Properties' => [
--- a/config/app.default.php
+++ b/config/app.default.php
@@ -515,6 +515,14 @@ return [
      *  - 'bulk' bulk actions list
      *  - 'fastCreate' fields for fast creation forms, by type
      *  - 'labels' to customize labels by properties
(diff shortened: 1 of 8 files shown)


config/app.default.php
```diff
@@ -514,6 +514,15 @@
...
+     *  - 'labels' to customize labels by properties
+     *    + 'fields'
+     *      + fieldName => string
+     *      + fieldName => string
+     *      + fieldName => string
+     *    + 'options'
+     *      + fieldName => array
+     *      + fieldName => array
+     *      + fieldName => array
      *
```
stefanorosanelli:
```suggestion
     *  - 'labels' to customize labels of properties names and their options, it can have two keys `fields` and `options`
         + 'fields' contains an associative array with property names as keys and labels as values 
         + 'options' contains an associative array with property names as keys and an array containing option items and the custom labels to use
```

config/app.default.php
```diff
@@ -576,6 +585,13 @@
         //         'required' => ['status', 'title'],
         //         'all' => ['status', 'title', 'description'],
         //     ],
+        //     'labels' => [
+        //         'status' => [
+        //             'on' => 'O N',
+        //             'off' => 'O F F',
+        //             'draft' => 'D R A F T',
+        //         ],
+        //     ],
```
stefanorosanelli:
```suggestion
        //     'labels' => [
        //         'fields' => [
        //             'description' => 'summary',
        //         ],
        //         'options' => [
        //             'status' => [
        //                 'on' => 'published',
        //                 'off' => 'archived',
        //                 'draft' => 'in progress',
        //             ],
        //         ],
        //     ],
```

stefanorosanelli (approved):

PR #780, opened 2022-05-20
Clone object with relations

didoda:
This provides the possibility to clone object with relations (excluding "children").

When you click on clone, prompt shows a new check "Clone relations". When checked, the cloning process copies relations from source object to new object.

Note: tests for `CloneComponent` use mocked api, because we found a bug (https://github.com/bedita/bedita/pull/1912) that does not allow us to create relations as we should to have a full test.

### DIFF
--- a/src/Controller/Component/CloneComponent.php
+++ b/src/Controller/Component/CloneComponent.php
@@ -71,11 +71,13 @@ class CloneComponent extends Component
      */
     public function filterRelations(array $relationships): array
     {
-        return array_filter(
-            $relationships,
-            function ($relationship) {
-                return !in_array($relationship, ['children', 'parents', 'translations']);
-            }
+        return array_values(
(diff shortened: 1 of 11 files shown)


src/Controller/Component/CloneComponent.php
```diff
@@ -0,0 +1,126 @@
...
+     * @param array $relationships The relationships
+     * @return array
+     */
+    public function filterRelations(array $relationships): array
+    {
+        return array_values(
+            array_filter(
+                $relationships,
+                function ($relationship) {
+                    return !in_array($relationship, ['children', 'parents', 'translations']);
```
stefanorosanelli: we could make `public` and use `ModulesComponent::FIXED_RELATIONSHIPS`

stefanorosanelli (approved):

PR #782, opened 2022-05-27
Refactor categories

didoda:
This refactor categories pages (in Modules view and Model).

This introduces `<category>` component to better handle UI logics, as:

 - category name already used
 - disable modify / create buttons, if category data is not valid
 - show error feedback, in a more readable way


src/Controller/AppController.php
```diff
@@ -32,6 +32,13 @@
  */
 class AppController extends Controller
 {
+    /**
+     * Object type currently used
+     *
+     * @var string
+     */
+    protected $objectType = null;
```
stefanorosanelli: we could probably add `$objectType` to `CategoriesController` like we did in `BulkController` and not define in `AppController` as it will be used by just 3 controllers.

src/Controller/Component/CategoriesComponent.php
```diff
@@ -45,6 +45,27 @@ public function index(?string $objectType = null, ?array $options = []): array
...
+     *
+     * @param string|null $objectType The object type
+     * @return array
+     */
+    public function names(?string $objectType = null): array
+    {
+        $apiClient = ApiClientProvider::getApiClient();
+        $query = [
+            'fields' => 'name',
+            'page_size' => 1000,
```
stefanorosanelli: in BE4 API we have a hard `MAX_LIMIT` of 500 - i don't think `1000` will be ever used 

it's ok for now, but we should add more calls if there are more pages

src/Controller/ModulesController.php
```diff
@@ -32,21 +32,13 @@
  */
 class ModulesController extends AppController
 {
-    /**
-     * Object type currently used
-     *
-     * @var string
-     */
-    protected $objectType = null;
```
stefanorosanelli: i'd prefer to restore `$objectType` - see before

stefanorosanelli (approved):

PR #783, opened 2022-05-31
Enhance Date Ranges, adding "Every day" and weekdays

didoda:
This introduces new features in `date-ranges`, by refactoring `date-ranges-view.js`:

 - handle `Every day` check and week days
 - save `dateRange.params` as json
 - fix `All day` behaviour
 - fix UI/UX

An example:

date-ranges

### DIFF
--- a/src/Locale/en_US/default.po
+++ b/src/Locale/en_US/default.po
@@ -1,7 +1,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: BEdita 4 \n"
-"POT-Creation-Date: 2022-05-06 08:14:16 \n"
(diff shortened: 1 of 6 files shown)


stefanorosanelli (commented):
When just `From` or `To` date items are selected or if `From` and `To` are on the same day the `Every day` check box is not significant: it's probably best to hide or disable it

didoda:
> When just `From` or `To` date items are selected or if `From` and `To` are on the same day the `Every day` check box is not significant: it's probably best to hide or disable it

good point. but "appearing" and "disappearing" checkbox could confuse the user. I'll give it a try...

PR #786, opened 2022-06-03
Pagination current page as input

didoda:
This provides a change to pagination block: current page is an editable input text.

This allows user to jump to a desired page, avoiding multiple clicks on prev or next button.

On input change, if value is not an integer or greater than pagination count, it's cleared.

It looks like follows:
pagination


stefanorosanelli (changes requested):
On objects index no problems, on related objects tab a page reload is performed and the input page value is lost (restored to 1)

stefanorosanelli (approved):

PR #789, opened 2022-06-08
Cakephp 4.4.0

didoda:
This updates `cakephp/cakephp` to `4.4.0` (ref: https://book.cakephp.org/4/en/appendices/4-4-migration-guide.html).

`cakephp/console` is added in `require-dev` (console classes are not in `cakephp/cakephp:4.4.0`).

Some files updates, from a comparison with https://github.com/cakephp/app : 

 - config/.env.example
 - config/app.default.php
 - config/bootstrap.php
 - config/bootstrap_cli.php
 - config/paths.php
 - src/Application.php


stefanorosanelli (approved):

config/bootstrap.php
```diff
@@ -132,9 +132,9 @@
  */
 $isCli = PHP_SAPI === 'cli';
 if ($isCli) {
-    (new ConsoleErrorHandler(Configure::read('Error')))->register();
+    (new ExceptionTrap(Configure::read('Error')))->register();
```
batopa: I think it would be set both for cli and not, see https://github.com/cakephp/app/blob/4.x/config/bootstrap.php#L125

I would do a double check against cakephp/app to see if other changes on main files (`boostrap.php`, `bootstrap-cli.php`, `app.php`, `paths.php', `Application.php` etc...) can be ported

batopa (changes requested):

batopa (approved):

Issue #803, opened 2022-07-04
Appearence configurations


didoda:
`Administration->Appearence` now provides management of `Modules` configuration.

We should provide more configurations:

- [x] `AlertMessage` (https://github.com/bedita/manager/pull/814)
- [x] `Export` (https://github.com/bedita/manager/pull/817)
- [x] `Modules` (https://github.com/bedita/manager/pull/799)
- [x] `Pagination` (https://github.com/bedita/manager/pull/817)
- [x] `Project` (https://github.com/bedita/manager/pull/814)
- [x] `Properties` (https://github.com/bedita/manager/pull/807)

didoda:
All tasks completed

PR #807, opened 2022-07-05
Admin Appearence - Properties configuration

didoda:
This provides a part of https://github.com/bedita/manager/issues/803, configuration `Properties`.

### DIFF
--- a/src/Controller/Admin/AppearenceController.php
+++ b/src/Controller/Admin/AppearenceController.php
@@ -71,14 +71,8 @@ class AppearenceController extends AdministrationBaseController
         $this->getRequest()->allowMethod(['post']);
         $data = (array)$this->getRequest()->getData();
         $propertyName = (string)Hash::get($data, 'property_name');
-        foreach ($this->properties as $property) {
-            if ($property !== $propertyName) {
-                continue;
-            }
-            $content = (string)Hash::get($data, Inflector::camelize($property));
-            $this->Config->save(Inflector::camelize($property), (array)json_decode($content, true));
-            break;
-        }
+        $content = (string)Hash::get($data, Inflector::camelize($propertyName));
+        $this->Config->save(Inflector::camelize($propertyName), (array)json_decode($content, true));
 
         return $this->redirect(['_name' => 'admin:list:appearence']);
     }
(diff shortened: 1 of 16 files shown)


src/Controller/Admin/AppearenceController.php
```diff
@@ -55,8 +70,15 @@ public function save(): ?Response
...
         $this->getRequest()->allowMethod(['post']);
         $data = (array)$this->getRequest()->getData();
-        $modules = (string)Hash::get($data, 'Modules');
-        $this->Config->save('Modules', (array)json_decode($modules, true));
+        $propertyName = (string)Hash::get($data, 'property_name');
+        foreach ($this->properties as $property) {
+            if ($property !== $propertyName) {
+                continue;
+            }
+            $content = (string)Hash::get($data, Inflector::camelize($property));
```
stefanorosanelli: we could skip the foreach and use just 

```php
$content = (string)Hash::get($data, Inflector::camelize($propertyName));
$this->Config->save(Inflector::camelize($propertyName), (array)json_decode($content, true));
```

src/Controller/Component/PropertiesComponent.php
```diff
@@ -95,9 +95,11 @@ public function startup(): void
 
             return;
         }
-
+        /** @var \App\Controller\ModulesController $controller */
```
stefanorosanelli: we could add the `Config` component to `PropertiesComponent` simply declaring

```php
    protected $components = ['Config'];
```
Then simply call `$this->Config` methods inside the component

src/Controller/Component/PropertiesComponent.php
```diff
@@ -95,9 +100,10 @@ public function startup(): void
 
             return;
         }
-
+        /** @phpstan-ignore-next-line */
+        $this->Config->read('Properties');
```
stefanorosanelli: I suggest this change to avoid too many cache hits and to avoid changes in `Control` and `PropertyHelper` classes 
In `ConfigComponent::read()` line 71 we could write: 
```php
        if (!empty($config)) {
            $content = (array)json_decode((string)Hash::get($config, 'attributes.content'), true);
            Configure::write($key, $content);
        }
``` 

This way here in `PropertiesComponent` we would have:

```php
Configure::load('properties');
$properties = $this->Config->read('Properties');
```

`Control` and `PropertyHelper` will keep using `Configure::read` and remain unchanged

Issue #805, opened 2022-07-04
Data Modeling Relation modify


didoda:
Provide following enhancements:

- [x] object types left and right as list of checkboxes + better UI (https://github.com/bedita/manager/pull/808)

didoda:
Solved with #808

PR #808, opened 2022-07-05
Model relations view object types lists

didoda:
This provides https://github.com/bedita/manager/issues/805, refactoring object types part in relation view.

### DIFF
--- a/src/Controller/Model/RelationsController.php
+++ b/src/Controller/Model/RelationsController.php
@@ -157,7 +157,10 @@ class RelationsController extends ModelBaseController
     protected function allTypes(): array
     {
         try {
-            $response = $this->apiClient->get('/model/object_types');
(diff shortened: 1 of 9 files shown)


src/Controller/Model/RelationsController.php
```diff
@@ -146,4 +148,23 @@ protected function relatedTypes(array $resource, string $side): array
...
+
+    /**
+     * Get all types
+     *
+     * @return array
+     */
+    protected function allTypes(): array
+    {
+        try {
+            $response = $this->apiClient->get('/model/object_types');
```
stefanorosanelli: we shoul add a query filter like
```php
[
    'page_size' => 100,
    'filter' => ['enabled' => true],
]
```
to avoid pagination issues and exclude disabled types or use `SchemaComponent::objectTypesFeatures()` to get object types data and parent relations

Issue #804, opened 2022-07-04
Data Modeling Object Type modify


didoda:
Provide the following enhancements:

- [x] Single custom property Hide/Show button should populate `Hidden` field properly (https://github.com/bedita/manager/pull/810)
- [x] Single custom property Delete button should make an ajax call to api delete proper endpoint (https://github.com/bedita/manager/pull/812)
- [x] Parent name should be an abstract objects only select combo (https://github.com/bedita/manager/pull/815)
- [x] Table should be a select combo (static list for now, result of an api call in the future) (https://github.com/bedita/manager/pull/815)
- [x] Associations should be a list of checkboxes (static list for now: DateRanges, Streams, Categories, Tags; there could be an extra value from api: show it as readonly checked) (https://github.com/bedita/manager/pull/816)
- [x] Single custom property Type should be a property_types select combo (https://github.com/bedita/manager/pull/816)

didoda:
All tasks completed

PR #815, opened 2022-07-08
Data Modeling -> Object types modify: parent_name and table as combo

didoda:
This deals with two tasks from https://github.com/bedita/manager/issues/804:

 - Parent name should be an abstract objects only select combo
 - Table should be a select combo (static list for now, result of an api call in the future)

A new configuration has been introduced: `Model.objectTypesTables`, used to fill the select combo list for `table` in data modeling (config list of tables is added to core tables). Configuration example:

```
'Model' => [
    'objectTypesTables' => [
        'MyPlugin.Cats',
        'MyPlugin.Dogs',
    ],
],
```

### DIFF
--- a/config/app_local.example.php
+++ b/config/app_local.example.php
@@ -408,11 +408,20 @@ return [
      */
     // 'Maintenance' => [
     //     'message' => 'This page won\'t be available for some time. Try later',
-    // ]
+    // ],
 
     /**
      * The manager application name in the API (default is `manager`).
      * This is used in reading/writing manager configuration data.
      */
     // 'ManagerAppName' => 'my-manager-app',
+
+    /**
+     * Object types to add to core types, used in Data Modeling -> Object types.
+     */
+    // 'Model' => [
+    //     'objectTypesTables' => [
+    //         'MyPlugin.MyTable',
+    //     ],
+    // ],
 ];
--- a/config/app_local.example.php
+++ b/config/app_local.example.php
@@ -417,7 +417,7 @@ return [
     // 'ManagerAppName' => 'my-manager-app',
 
     /**
-     * Object types to add to core types, used in Data Modeling -> Object types.
+     * Object types to add to core tables, used in Data Modeling -> Object types.
      */
     // 'Model' => [
     //     'objectTypesTables' => [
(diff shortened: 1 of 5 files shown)


src/Controller/Model/ObjectTypesController.php
```diff
@@ -24,6 +25,51 @@
  */
 class ObjectTypesController extends ModelBaseController
 {
+    /**
+     * Core tables list.
+     *
+     * @var array
+     */
+    public const TABLES = [
+        'BEdita/Core.Annotations',
```
stefanorosanelli: actual core object tables are just

```php
[
   'BEdita/Core.Folders',
   'BEdita/Core.Links',
   'BEdita/Core.Locations',
   'BEdita/Core.Media',
   'BEdita/Core.Objects',
   'BEdita/Core.Profiles',
   'BEdita/Core.Publications',
   'BEdita/Core.Users'
]
```

src/Controller/Model/ObjectTypesController.php
```diff
@@ -56,12 +102,58 @@ public function view($id): ?Response
...
+    }
+
+    /**
+     * Get abstract types from api
+     *
+     * @return array
+     */
+    protected function abstractTypes(): array
+    {
+        $response = [];
```
stefanorosanelli: Simpler version
```php
$features = $this->Schema->objectTypesFeatures();
$types = array_keys($features['descendants'));
sort($types);
```

we could also create a new method `SchemaComponent::abstractTypes()` with this logic

config/app_local.example.php
```diff
@@ -408,11 +408,20 @@
...
+    // ],
 
     /**
      * The manager application name in the API (default is `manager`).
      * This is used in reading/writing manager configuration data.
      */
     // 'ManagerAppName' => 'my-manager-app',
+
+    /**
+     * Object types to add to core types, used in Data Modeling -> Object types.
```
stefanorosanelli: [minor]
```Object tables to add to core tables,...```

stefanorosanelli (commented):
* `Table` and `Parent Name` on core types should be read only, as in previous version
* `Parent Name` can be NULL, only for `Objects` usually, but it's a possible value in general
*  if a `Table` is not present in configuration but set in an object type it should be displayed

stefanorosanelli (approved):

PR #816, opened 2022-07-11
Data Modeling -> Object types modify: custom property type as combo + associations as list of checkboxes

didoda:
This provides 2 tasks from https://github.com/bedita/manager/issues/804:

- Single custom property Type should be a property_types select combo
- Associations should be a list of checkboxes (static list for now: DateRanges, Streams, Categories, Tags; there could be an extra value from api: show it as readonly checked)

### DIFF
--- a/src/Controller/Component/PropertiesComponent.php
+++ b/src/Controller/Component/PropertiesComponent.php
@@ -312,7 +312,7 @@ class PropertiesComponent extends Component
     /**
      * Get associations options for select multiple as list of checkboxes
      *
-     * @param array $value
+     * @param array $value The value
      * @return array
      */
     public function associationsOptions(array $value): array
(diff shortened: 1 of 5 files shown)


src/Controller/Component/PropertiesComponent.php
```diff
@@ -26,6 +27,23 @@
  */
 class PropertiesComponent extends Component
 {
+    /**
+     * Schema property types
+     */
+    public const CUSTOM_PROPERTY_TYPES = [
```
stefanorosanelli: We should read `property_types` via `GET /model/property_types` to include `core` and `custom` types.

It would be nice to see the JSON-SCHEMA once a property is selected (but we can to this later)

stefanorosanelli (approved):

PR #825, opened 2022-07-26
Tag picker tag label and name

didoda:
This provides an enhancement in tag picker: show tag label followed by tag name between square brackets. It can be handy when user wants to know the exact unique name of a tag.

An example:

tag-picker


qwerg:
Per me OK.

PR #826, opened 2022-07-27
Admin Applications client_secret

didoda:
This adds `client_secret` to the "secrets" in `Admin => Applications` page.


stefanorosanelli (commented):
-  I don't see buttons to create/change/remove applications
- if `client_secret` is NULL probably something like `---` or a blank field is preferrable 

image

PR #830, opened 2022-08-09
Date range validation

didoda:
This introduces a javascript validation on date ranges, to avoid saving wrong data.


qwerg:
Some notes:
- `empty date` should not be an error or warning
- `end date` it’s not mandatory
- it’s not necessary to show the `OK` validation

In short, I think we should report only the `invalid date` error (`date end` before `date start`)

Graphics:
I would draw the error report like this, strengthening the date field (red) and reporting only the error, not the OK
screen

qwerg:
Side note: can we show the time format without the seconds?
They can't be set from the date panel anyway...

didoda:
@qwerg just pushed for:

 - no error/warning on empty dates
 - red background on invalid date range interval
 - no seconds in dates

stefanorosanelli (approved):

PR #839, opened 2022-09-06
Introduce BEdita v5 in CI

stefanorosanelli:


.github/workflows/php.yml
```diff
@@ -13,5 +13,24 @@ on:
...
+
+  stan:
+    uses: bedita/github-workflows/.github/workflows/php-stan.yml@v1
+    with:
+      php_versions: '["7.4", "8.0", "8.1"]'
+
+  unit-4:
+    uses: bedita/github-workflows/.github/workflows/php-unit.yml@v1
+    with:
+      php_versions: '["7.4"]'
```
didoda:
```suggestion
      php_versions: '["7.4","8.0","8.1"]'
```
it takes more time, but it make sense to test against the 3 php versions

didoda (changes requested):

didoda (approved):

Issue #842, opened 2022-09-08
Admin - Roles and modules access


didoda:
BEM provides modules accesses per role with the configuration `AccessControl`:

An example:
```
'AccessControl' => [
     'manager' => [
         'hidden' => ['objects'],
         'readonly' => ['documents'],
     ],
     'guest' => [
         'hidden' => ['objects', 'users'],
         'readonly' => ['documents'],
     ],
],
```

Admin users should be able to change `AccessControl` through an an `Admin - Roles and modules access` page, with an UI like the following:

https://xd.adobe.com/view/48d5c8b0-ea5e-4199-9165-591641d31b3f-1d86/screen/f498cac4-19db-4b0e-82b4-957e5e39eec8/

The available roles should be filtered considering permissions for roles on `/auth` endpoint for `manager` application (if any).

Note: reuse the pattern of overriding the configuration in `configs` via proper API calls.

didoda:
Done with #844

PR #844, opened 2022-09-12
Roles Modules accesses admin page

didoda:
This provides https://github.com/bedita/manager/issues/842, Roles Modules accesses admin page

### DIFF
--- a/src/Template/Element/Admin/sidebar.twig
+++ b/src/Template/Element/Admin/sidebar.twig
@@ -1,5 +1,5 @@
 {# Append urls to sidebar #}
-{% set actions = ['appearence', 'applications', 'endpoints', 'config', 'async_jobs', 'roles', 'endpoint_permissions'] %}
+{% set actions = ['appearence', 'applications', 'endpoints', 'config', 'async_jobs', 'roles', 'roles_modules', 'endpoint_permissions'] %}
 {% set controllerAction = _view.request.getparam('controller')|default('') %}
 {% for action in actions %}
     {% if action == controllerAction|underscore %}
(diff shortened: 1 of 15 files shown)


stefanorosanelli (changes requested):
After a quick test:

* `admin` should be excluded in roles list, setting permissions is not effective anyway 

* roles not able to access the Manager app should also be excluded, maybe we could use a configuration for that or better you could find out roles looking at endpoint permissions on `/auth` endpoint, maybe not that simple  

* we should show if some modules-endpoints are already filtered via endpoint permissions (not easy, we can do this later), otherwise some permission here could be not effective (e.g. a module already readonly for a role via endpoint permissions) 

* in roles list we could simply display `manager` instead `manager 2 Manager Role` 

* when a module is read only (via endpoint permissions or manager ACL) an icon or some visual element coudl be useful (TODO in other issue)

src/Template/Element/Admin/sidebar.twig
```diff
@@ -1,5 +1,5 @@
 {# Append urls to sidebar #}
-{% set actions = ['appearence', 'applications', 'endpoints', 'config', 'async_jobs', 'roles', 'endpoint_permissions'] %}
+{% set actions = ['appearence', 'applications', 'endpoints', 'config', 'async_jobs', 'roles', 'roles_modules', 'endpoint_permissions'] %}
```
stefanorosanelli, line 2: Probably a better link label is `Modules Access` or `Modules ACL` - we can chenge it later

stefanorosanelli (approved):

src/Template/Element/Admin/sidebar.twig
didoda: We'll do it later

PR #847, opened 2022-09-15
Default lang for combo language in object view

didoda:
This fixes a buggy behaviour in object view (new or edit):

 - (new) first language is selected, not the default one (config `Project.config.I18n.default`).
 - (edit) when object.attributes.lang is null or empty, first language is selected

Expected behaviour:

 - (new) default language is selected ('en' when there's no config `Project.config.I18n.default`)
 - (edit) when object.attributes.lang is null or empty, show empty option; user can change it to proper language

Note: in "edit" case, it's dangerous to set a default language... objects with "null" language would be treated as they have a language. This could confuse the user

Bonus: fix `ApiConfigTrait` and `ApiConfigTraitTest` (phpstan found some issues)

### DIFF
--- a/src/Template/Element/Form/related_translations.twig
+++ b/src/Template/Element/Form/related_translations.twig
@@ -17,7 +17,7 @@
             <div v-show="isOpen" class="tab-container">
 
                 <div>
-                    {% set lang = object.id ? object.attributes.lang : config('I18n.default')|default('en') %}
+                    {% set lang = object.id ? object.attributes.lang : config('Project.config.I18n.default') %}
                     {{ Property.control('lang', lang, { 'label': __('The main language is') })|raw }}
 
                     {% if object.relationships.translations %}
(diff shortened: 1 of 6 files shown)


src/Template/Element/Form/related_translations.twig
```diff
@@ -17,7 +17,8 @@
             <div v-show="isOpen" class="tab-container">
 
                 <div>
-                    {{ Property.control('lang', object.attributes.lang, { 'label': __('The main language is') })|raw }}
+                    {% set lang = object.id ? object.attributes.lang : config('I18n.default')|default('en') %}
```
stefanorosanelli: We should use `Project.config.I18n.default` here and not `I18n.default`

See `App\Form\Options::lang()`

`I18n` config key refers to the Manager UI languages and `Project.config.I18n` refers to the object languages - the two sets can be different. 

Anyway I would remove `|default('en')` here
It should be:
```
{% set lang = object.id ? object.attributes.lang : config('Project.config.I18n.default') %}
```

stefanorosanelli (changes requested):

stefanorosanelli (approved):

PR #852, opened 2022-09-28
Fast create form: fields types

### LINKED ISSUE #850: Error on `fastCreate` with JSON property
On a custom [`fastCreate` configuration](https://github.com/bedita/manager/wiki/Setup:-Properties-display#fastcreate) if one of the properties is a JSON property object creation fails.

From the JS error console we get:

```
vue.js:1906 TypeError: Cannot read properties of undefined (reading 'id')
    at VueComponent.selectClasses (relations-add.js:347:1)
    at eval (eval at createFunction (vue.js:11698:1), <anonymous>:48:41516)
```

didoda:
This fixes #850 (a bug) and  provides a new feature: `fastCreate` form can show proper fields (textarea, input, etc.), according to the `fastCreate` configurations.

If no configuration is set, the system uses objects schemas to decide proper fields types: this PR provides a fix in `/:object_type/view/new`, by loading schemas for relatable objects.

Config example:
```
'Properties'
    documents' => [
        'fastCreate' => [
            'required' => ['status'],
            'all' => [
                'status',
                'title',
                'description' => 'plaintext',
                'body' => 'richtext',
                'extra' => 'json',
            ],
        ],
    ],
],
```

This config allows you to override the schema type, to customize fast create object form; a "richtext" could be represented with a "plaintext", for UI reasons

When json fields are involved, this adds a `_jsonKeys` hidden field (the patch for #850).

Backward compatibility with previous type of config.


src/View/Helper/PropertyHelper.php
```diff
@@ -152,4 +154,112 @@ public function value(array $resource, string $property): string
...
+     */
+    public function fastCreateFields(string $type, string $prefix): string
+    {
+        $cfg = (array)Configure::read(sprintf('Properties.%s.fastCreate', $type));
+        $fields = (array)Hash::get($cfg, 'all', ['status', 'title', 'description']);
+        $required = (array)Hash::get($cfg, 'required', ['status', 'title']);
+        $html = '';
+        $jsonKeys = [];
+        $ff = [];
+        foreach ($fields as $field => $fieldType) {
```
stefanorosanelli, line 173: Using this `foreach` ti works for me


```php
        foreach ($fields as $key => $val) {
            $field = $key;
            $fieldType = $val;
            if (is_numeric($key)) {
                $field = $val;
                $fieldType = ControlType::fromSchema($this->schema($field, $type));
            }
            $fieldClass = !in_array($field, $required) ? 'fastCreateField' : 'fastCreateField required';
            $fieldOptions = [
                'id' => sprintf('%s%s', $prefix, $field),
                'class' => $fieldClass,
                'data-name' => $field,
                'key' => sprintf('%s-%s', $type, $field),
            ];
            if ($field === 'date_ranges') {
                $html .= $this->dateRange($type, $fieldOptions);
                continue;
            }
            if ($fieldType === 'json') {
                $jsonKeys[] = $field;
            }
            // override field options if we use a custom type
            if (!is_numeric($key)) {
                $this->prepareFieldOptions($field, $fieldType, $fieldOptions);
            }
            $html .= $this->control($field, '', $fieldOptions, $type);
            $ff[] = $field;
        }
```

didoda: This should not be necessary, because `$this->control` internally uses object schema, if available.

stefanorosanelli (approved):

PR #853, opened 2022-09-28
Introduce custom bulk actions

stefanorosanelli:
This PR introduces custom bulk actions.
These bulk actions can be configured in `Properties.{objectType}.bulk` key like this:

```php
   'bullk' => [
      'status', // property bulk action, without label
      'Change language' => 'lang', // property bulk action, with label 
      'Action Label' => '\MyApp\BulkActionClass', // custom class to perform bulk action, with label
   ],
```

When a bulk action item in the `bulk` array does not match a property name it will be treated like a custom bulk action. 
Custom bulk action item must be the name of a class implementing `CustomBulkActionInterface` interface. Its `bulkAction()` method is then called to perform the desired object manipulation.

### DIFF
--- a/src/Template/Element/Form/bulk_custom.twig
+++ b/src/Template/Element/Form/bulk_custom.twig
@@ -23,6 +23,6 @@
             </label>
             <button id="custom-bulk-{{ key }}" class="button button-outlined" @click.prevent="bulkActions('{{ bulkFormId }}')" :disabled="!selectedRows.length">{{ __('Ok') }}</button>
         </div>
-    {{ Form.end()|raw }}
+        {{ Form.end()|raw }}
     {% endif %}
 {% endfor %}
--- a/src/Template/Element/Form/bulk_custom.twig
+++ b/src/Template/Element/Form/bulk_custom.twig
@@ -1,0 +1,28 @@
+{# bulkActions set in config Properties.<type>.bulk #}
+{% for label, key in bulkActions %}
+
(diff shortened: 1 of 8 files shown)


src/Template/Element/Form/bulk_custom.twig
```diff
@@ -0,0 +1,28 @@
...
+            <label for="custom-bulk-{{ key }}">
+                {%- if label matches '/^\\d+$/' -%}
+                    {{ __(key|humanize) }}
+                {%- else -%}
+                    {{ __(label|humanize) }}
+                {%- endif -%}
+            </label>
+            <button id="custom-bulk-{{ key }}" class="button button-outlined" @click.prevent="bulkActions('{{ bulkFormId }}')" :disabled="!selectedRows.length">{{ __('Ok') }}</button>
+        </div>
+    {{ Form.end()|raw }}
```
didoda: minor: indent

src/Template/Element/Form/bulk_properties.twig
```diff
@@ -41,4 +43,5 @@
             <button class="button button-outlined" @click.prevent="bulkActions('bulk-{{ key }}')" :disabled="!selectedRows.length">{{ __('Ok') }}</button>
         </div>
     {{ Form.end()|raw }}
```
didoda: minor: indent as ``{{ Form.create`

didoda (approved):

PR #872, opened 2022-10-25
bedita image 5.0.10

didoda:


didoda:
Not necessary. Update will be done with other changes

PR #879, opened 2022-11-02
Tags module

didoda:
This provides "tags" module among other "user" modules: same logic of "Modelling / Tags", with new UI/UX.

tags

### DIFF
--- a/src/Controller/Component/ModulesComponent.php
+++ b/src/Controller/Component/ModulesComponent.php
@@ -68,9 +68,7 @@ class ModulesComponent extends Component
      * @var array
      */
     protected $otherModules = [
-        'tags' => [
-            'color' => '#6F12B7',
-        ],
+        'tags',
     ];
 
     /**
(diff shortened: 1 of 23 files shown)


src/Controller/Component/ModulesComponent.php
```diff
@@ -62,6 +62,17 @@ class ModulesComponent extends Component
      */
     protected $modules = [];
 
+    /**
+     * Other "logic" modules, non objects
+     *
+     * @var array
+     */
+    protected $otherModules = [
```
stefanorosanelli, line 70: Suggested change:

1. add `tags` module with color in default `config/app.php` 
2. in `$otherModules` list only module names, like 
```php 
protected $otherModules = [
    'tags', 
];
```
3. in `getModules()` change  
 ```$metaModules = $this->modulesFromMeta();```
to 
 ``` $metaModules = $this->modulesFromMeta() + array_flip($this->otherModules);```
4. remove `$this->otherModules` from `array_merge` below

stefanorosanelli (approved):

src/Controller/Component/ModulesComponent.php
didoda: ok. but this way the color is not preserved, because `Modules` config from `config/app.php` is almost always overwritten

PR #880, opened 2022-11-03
Fix json editor save

### LINKED ISSUE #877: Json field save
Json field save with "empty" values is buggy.

# Expected behaviour

When value is empty string, data is saved as null / json editor shows null in the json editor.
 
# Actual behaviour

When value is empty string, API error is thrown (when field is a custom property), and the value is not saved.

didoda:
This fixes https://github.com/bedita/manager/issues/877

### DIFF
--- a/resources/js/app/components/json-editor-input.js
+++ b/resources/js/app/components/json-editor-input.js
@@ -21,42 +21,36 @@ export default {
     },
 
     async mounted() {
-        const element = this.el;
-        const content = element.value;
         try {
-            const json = content !== "" && JSON.parse(content) || {};
-
-            if (json) {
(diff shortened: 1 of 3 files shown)


stefanorosanelli (approved):
Every use case seems to be handled correctly now.

The only problem is when a JSON field is an empty object `{}` it is displayed as an empty array `[]`.  But this is how it is returned by the API - has to be fixed there.

stefanorosanelli (approved):
Every use case seems to be handled correctly now.

The only problem is when a JSON field is an empty object `{}` it is displayed as an empty array `[]`.  But this is how it is returned by the API - has to be fixed there.

PR #887, opened 2022-11-08
Meta, title and BEdita references updates

didoda:
This provides some minor updates:

 - remove unused meta tags
 - add some meta tags
 - replace "BEdita 4" with "BEdita Manager" (when the reference was wrong)
 - replace "BEdita 4" with "BEdita", when API is considered


templates/Layout/default.twig
```diff
@@ -1,28 +1,33 @@
...
 
 <html lang="{{ config('I18n.lang', 'en')}}">
 <head>
     {{ Html.charset()|raw }}
-    {{ element('meta') }}
-    <title>{{ "#{_view.fetch('title')} | #{project.name ?: 'BEdita 4'}" }}</title>
+
+    {{ Html.meta('viewport', 'width=device-width, initial-scale=1.0')|raw }}
+    {{ Html.meta('description', 'BEdita Manager')|raw }}
+    {{ Html.meta({
```
stefanorosanelli: we could skip this `meta rel=manifest` - I don't think it applies in our case https://developer.mozilla.org/en-US/docs/Web/Manifest

templates/Element/meta.twig
```diff
@@ -1,25 +0,0 @@
-<meta name="viewport" content="width=device-width, initial-scale=1.0">
-<meta name="description" content="">
```
stefanorosanelli, line 2: we could proably keep `description` and add `generator` ?

```
<meta name="description" content="BEdita Manager, official BEdita API admin tool">
<meta name="generator" content="BEdita Manager v#.#.#">
```

stefanorosanelli (approved):

PR #915, opened 2023-01-13
Icons on buttons and anchors

didoda:
This provides an update in twig templates to use proper icons on buttons and anchors.


qwerg:
Ok

PR #916, opened 2023-01-20
File upload: check file type and size. Show preview if it is an image

### LINKED ISSUE #922: File upload: preview (image), check file type and size, title
When uploading a file, there could be some improvements in UI/UX and in the behaviour.
File upload is in:

 - create new media page
 - edit media page
 - fast create media (from side panel)
 - bulk upload media from drag and drop in relation section

Expected behaviour
===============

During upload, when a file is choosen, there should be a check on file mime type and filesize (modal warn, if something is not fine). If everything is fine, a figure preview should appear (for images). When the page is "create new media" or "edit media", if title is empty, it should have the value of the file name (without special characters and extension).

didoda:
This resolves https://github.com/bedita/manager/issues/922, an enhancement in Media view and other parts where upload file is involved.
When choosing a file, a check on type is performed: if file type is not among accepted mime types, the upload is not performed and the user is notified about it.
Two check are performed on save:

- check if file mime type is fine
- check if the file size is fine

If it's an image, a `figure` with the preview image is shown, on file change.
Data is saved on `Save` button click (as it was before this PR).

If media title is empty, on preview update the title value is populated with the file name (with some string replacement, to make it more friendly).

Preview + check file size + check file extension (for images) in:

 - media view for new object, upload file section
 - media view for edit object, upload file section
 - fast create form (side panel), upload file section

Accepted mime types by object type (this could be updated in future):

```
images: [
    'image/apng',
    'image/bmp',
    'image/jp2',
    'image/jpeg',
    'image/jpg',
    'image/gif',
    'image/png',
    'image/svg+xml',
    'image/webp',
],
```

### DIFF
--- a/resources/js/app/helpers/view.js
+++ b/resources/js/app/helpers/view.js
@@ -128,6 +128,26 @@ export default {
 
                 return title.trim();
             },
+
+            setTitleFromFileName(titleId, fileName) {
+                const elem = document.getElementById(titleId);
+                if (!elem || elem.value !== '') {
+                    return;
+                }
+                elem.value = this.titleFromFileName(fileName);
+            },
+
+            updatePreviewImage(file, titleId, thumb) {
+                if (file?.name) {
+                    if (titleId) {
+                        this.setTitleFromFileName(titleId, file.name);
+                    }
+
+                    return window.URL.createObjectURL(file);
+                }
+
+                return thumb || null;
+            },
         }
     }
 };
--- a/resources/js/app/helpers/view.js
+++ b/resources/js/app/helpers/view.js
@@ -117,6 +117,17 @@ export default {
 
                 return true;
             },
+
+            titleFromFileName(filename) {
+                let title = filename;
+                title = title.replaceAll('-', ' ');
+                title = title.replaceAll('_', ' ');
+                if (title.lastIndexOf('.') > 0) {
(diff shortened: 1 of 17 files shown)


resources/js/app/helpers/view.js
```diff
@@ -92,6 +92,63 @@ export default {
                 return true;
             },
 
+            checkMimeForUpload(file, objectType) {
+                /** accepted mime types by object type for file upload */
+                const mimes = {
```
batopa: Maybe as improvement we could define a list of default allowed mime types in configuration and add the list of allowed mime types in `accept` property of  input file 
```html
<input type="file" accept="image/png,image/jpeg,...">
```

batopa:
I've found a bug in fast create action. Step to reproduce in "add objects" panel:
1. click on "Create new for `<relation>`
2. selecting `files` type
3. choose `pdf` file (for example)
4. change the type selection in `images` (the title will be reset but not input file)
5. write a title
6. click on "create" button

An image with a pdf file as stream will be created skipping the mime type validation.
I attach a litte video to better explain


bug_check_mime_type

resources/js/app/helpers/view.js
didoda: added a config `uploadMimeTypes`

batopa (approved):

PR #948, opened 2023-02-16
Fix control options for `Properties.<type>.options`

didoda:
This fixes an unexpected behaviour with config `Properties.<type>.options`: "readonly" not applied on some types (`text` and others).

### DIFF
--- a/src/View/Helper/SchemaHelper.php
+++ b/src/View/Helper/SchemaHelper.php
@@ -61,28 +61,24 @@ class SchemaHelper extends Helper
     public function controlOptions(string $name, $value, $schema = []): array
     {
         $options = Options::customControl($name, $value);
+        $objectType = (string)$this->_View->get('objectType');
+        $ctrlOptionsPath = sprintf('Properties.%s.options.%s', $objectType, $name);
+        $ctrlOptions = (array)Configure::read($ctrlOptionsPath);
+
         if (!empty($options)) {
-            $objectType = (string)$this->_View->get('objectType');
-            $ctrlOptionsPath = sprintf('Properties.%s.options.%s', $objectType, $name);
-            $ctrlOptions = (array)Configure::read($ctrlOptionsPath);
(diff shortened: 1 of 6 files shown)


src/View/Helper/SchemaHelper.php
```diff
@@ -62,7 +62,16 @@ public function controlOptions(string $name, $value, $schema = []): array
     {
         $options = Options::customControl($name, $value);
```
stefanorosanelli, line 63: I suggest this refactor on `controlOptions` method:

```php
    public function controlOptions(string $name, $value, $schema = []): array
    {
        $options = Options::customControl($name, $value);
        $objectType = (string)$this->_View->get('objectType');
        $ctrlOptionsPath = sprintf('Properties.%s.options.%s', $objectType, $name);
        $ctrlOptions = (array)Configure::read($ctrlOptionsPath);

        if (!empty($options)) {
            return array_merge($options, [
                'label' => Hash::get($ctrlOptions, 'label'),
                'readonly' => Hash::get($ctrlOptions, 'readonly', false),
                'disabled' => Hash::get($ctrlOptions, 'readonly', false),
            ]);
        }
        // verify if there's an handler by $type.$name
        if (!empty(Configure::read(sprintf('Control.handlers.%s.%s', $objectType, $name)))) {
            $handler = Configure::read(sprintf('Control.handlers.%s.%s', $objectType, $name));

            return call_user_func_array([$handler['class'], $handler['method']], [$value, $schema]);
        }
        $type = ControlType::fromSchema((array)$schema);

        return Control::control([
            'objectType' => $objectType,
            'property' => $name,
            'value' => $value,
            'schema' => (array)$schema,
            'propertyType' => Hash::get($ctrlOptions, 'type', $type),
            'label' => Hash::get($ctrlOptions, 'label'),
            'readonly' => Hash::get($ctrlOptions, 'readonly', false),
            'disabled' => Hash::get($ctrlOptions, 'readonly', false),
        ]);
    }
```

stefanorosanelli (approved):

PR #970, opened 2023-03-15
Fix module accesses on multiple roles

### LINKED ISSUE #969: Merge module accesses by role
Actual behaviour
=============

When two roles have different access type on the same module, the access on the module by a user that has both roles doesn't work as expected.
Example: the user "gustavo" belongs to two roles, "editor" and "guest" that have respectively "write access" and "read access" to the module "books". Gustavo has "readonly" access to the module "books".

Expected behaviour
===============

When two roles have different access type on the same module, the access on the module by a user that has both roles should be "the powerest" access of the two roles.
Example: the user "gustavo" belongs to two roles, "editor" and "guest" that have respectively "write access" and "read access" to the module "books". Gustavo should have "write" access to the module "books".

Briefly: when a user belongs to more roles, module accesses should be merged, considering the higher power role on each module.

Note: accesses can be `write`, `read`, `hidden`. The "power" scale is exactly `write > read > hidden`.

didoda:
This fixes https://github.com/bedita/manager/issues/969 by considering permissions priorities on modules ("write" > "read" > "hidden").

### DIFF
--- a/src/Controller/Component/ModulesComponent.php
+++ b/src/Controller/Component/ModulesComponent.php
@@ -200,16 +200,17 @@ class ModulesComponent extends Component
             $readonly = empty($readonly) ? $r : array_intersect($readonly, $r);
             $write = array_unique(array_merge($write, array_diff($modules, $hidden, $readonly)));
         }
-        $hidden = array_filter(
-            $hidden,
-            function ($item) use ($write) {
-                return !in_array($item, $write);
-            }
-        );
+        // Note: https://github.com/bedita/manager/issues/969 Accesses priority is "write" > "read" > "hidden"
         $readonly = array_filter(
             $readonly,
             function ($item) use ($write) {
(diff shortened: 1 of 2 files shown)


src/Controller/Component/ModulesComponent.php
```diff
@@ -189,14 +189,30 @@ protected function modulesByAccessControl(): void
...
+            function ($item) use ($write) {
+                return !in_array($item, $write);
+            }
+        );
+        $hidden = array_filter(
+            $hidden,
+            function ($item) use ($readonly, $write) {
+                return !in_array($item, $readonly) && !in_array($item, $write);
+            }
+        );
```
stefanorosanelli:
```suggestion
        $readonly = array_diff($readonly, $write);
        $hidden = array_diff($hidden, $readonly, $write);
```

stefanorosanelli (approved):

PR #972, opened 2023-03-17
Fix bulk operations in abstract object types modules

didoda:
This fixes some bugs in "bulk" operations on abstract object types modules:

 - filter status proper view
 - add loader on bulk button click
 - fix copy/move to folder (there was an error, when using abstract types)

### DIFF
--- a/src/Controller/BulkController.php
+++ b/src/Controller/BulkController.php
@@ -32,6 +32,13 @@ class BulkController extends AppController
     protected $objectType = null;
 
     /**
+     * Object type is abstract
+     *
+     * @var string
+     */
+    protected $abstractType = null;
+
+    /**
      * Selected objects IDs
      *
      * @var array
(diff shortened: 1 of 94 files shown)


src/Controller/BulkController.php
```diff
@@ -308,4 +308,22 @@ public function getErrors(): array
...
+    /**
+     * Get object type, when it is abstract
+     *
+     * @param string $id The object ID
+     * @param string $type The object type
+     * @return string
+     */
+    protected function getType(string $id, string $type): string
+    {
+        $schema = $this->Schema->getSchema($type);
```
stefanorosanelli: We don't need the second argument `$type` since `$this->objectType` is always used

We could also use `$this->Schema->abstractTypes()` in `initialize` method to see if `$this->objectType` is abstract just once and use a class attribute like `$abstractType` 

```php
protected function getType(string $id): string
{
    if (!$this->abstractType) {
        return $this->objectType;
     }
     // retrieve type via API 
}
 ```

PR #973, opened 2023-03-20
Use serial mode upload for dragndrop files to avoid uname error

paolosantarsiero:
Introducing serial mode upload for file with same basename in drag and drop area to avoid uname media error.

Ref: https://chialab.atlassian.net/browse/BEM-122

### DIFF
--- a/resources/js/app/components/drop-upload.js
+++ b/resources/js/app/components/drop-upload.js
@@ -105,7 +105,7 @@ export default {
             this.$el.classList.remove('dragover');
             // show all files in view
             [...files].forEach(file => this.setProgressInfo(file));
-            const [uniqueFiles, duplicatesFiles] = this.filterFiles(files);
+            const [uniqueFiles, duplicatesFiles] = this.filterFiles([...files]);
             uniqueFiles.forEach((file) => {
                 // skip file not allowed by size and remove it from view
                 if (this.$helpers.checkMaxFileSize(file) === false) {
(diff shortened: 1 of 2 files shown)


resources/js/app/components/drop-upload.js
```diff
@@ -101,19 +101,32 @@ export default {
...
 
+                this.upload(file).then((object) => this.uploadSuccessful(file, object));
             });
+
+            // use queue for file with same name
+            for (const file of duplicatesFiles) {
+                // skip file not allowed by size and remove it from view
+                if (this.$helpers.checkMaxFileSize(file) === false) {
+                    this.removeProgressItem(file);
+                    return;
```
didoda: I expected a `continue` here... otherwise this exits the function and other files in `duplicatedFiles` are not processed for upload

resources/js/app/components/drop-upload.js
```diff
@@ -101,19 +101,32 @@ export default {
...
+                    this.removeProgressItem(file);
                     return;
                 }
-                this.upload(f)
-                    .then((object) => {
-                        this.$emit('new-relations', [object]);
-                        this.removeProgressItem(f);
-                    });
 
+                this.upload(file).then((object) => this.uploadSuccessful(file, object));
```
didoda, line 116: `return` at line 113 should be `continue`?

resources/js/app/components/drop-upload.js
```diff
@@ -196,6 +209,28 @@ export default {
             this.$forceUpdate();
         },
 
+        uploadSuccessful(file, object) {
+            this.$emit('new-relations', [object]);
+            this.removeProgressItem(file);
+        },
+
+        getBaseNameFile(file) {
```
didoda: this function could be in `resources/js/app/helpers/view.js` with other "files" functions, as `checkMaxFileSize`; in this way it could be reused by `this.$helpers.getBaseNameFile(file)`

resources/js/app/components/drop-upload.js
paolosantarsiero: `return` in a `forEach()` callback is equivalent to continue

didoda (approved):

PR #978, opened 2023-03-30
Folders permissions

didoda:
This provides UI and logic changes to deal with folder permissions.

This requires a specific API versions (still WIP, https://github.com/bedita/bedita/pull/2000).

This introduces:

 - handle new association `Permissions` in object type modelling view: permissions logic is enabled (in API) only when object type has association `Permissions`
 - icons lock and unlock in tree-view specifying whether authenticated user has or not perms on folder
 - in folder view: "Permissions" section with a message saying if user has perms on folder, and listing roles as checkboxes
 - in folder view: when user has not permission, some buttons are not available (Save, Clone, Trash)

Bonus: refactor `.vue` files to properly handle `ttag` translations, see https://github.com/ttag-org/ttag/issues/201 for details (waiting for https://github.com/ttag-org/ttag-cli/pull/148 to be approved and eventually merged... a solution that would add support to use `t` inside `<template>`).

### DIFF
--- a/resources/js/app/components/permission/permission.vue
+++ b/resources/js/app/components/permission/permission.vue
@@ -40,7 +40,7 @@ export default {
         help() {
             const inherit = this.inherited ? this.msgInherited : '';
 
-            return `${msgYourRoles}: ${this.userRoles.join(',')} ${inherit}`;
+            return `${this.msgYourRoles}: ${this.userRoles.join(',')} ${inherit}`;
         },
         isLocked() {
             if (this.userRoles.includes('admin')) {
--- a/resources/js/app/components/permission/permission.vue
+++ b/resources/js/app/components/permission/permission.vue
@@ -1,7 +1,7 @@
 <template>
     <span v-if="hasPermission(role)" v-title="help()">
-        <Icon icon="carbon:locked" v-if="isLocked(role)"></Icon>
-        <Icon icon="carbon:unlocked" v-if="!isLocked(role)"></Icon>
+        <Icon icon="carbon:locked" v-if="isLocked()"></Icon>
+        <Icon icon="carbon:unlocked" v-if="!isLocked()"></Icon>
     </span>
 </template>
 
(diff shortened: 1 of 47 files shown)


resources/js/app/components/permission/permission.vue
```diff
@@ -0,0 +1,41 @@
...
+
+export default {
+
+    name: 'permission',
+
+    props: {
+        inherited: false,
+        objectRoles: [],
+        role: '',
+        userRoles: [],
```
le0m:
```suggestion
        inherited: {
            type: Boolean,
            default: false,
        },
        objectRoles: {
            type: Array,
            default: () => ([]),
        },
        role: {
            type: String,
            default: '',
        },
        userRoles: {
            type: Array,
            default: () => ([]),
        },
```

When defining `props` as an object, keys are prop name and values are prop types; if we want to define defaults each prop must be defined as an object with `type` and `default` properties.
Furthermore, defaults of props that are array or objects must be returned from a factory function.

resources/js/app/components/permission/permission.vue
```diff
@@ -0,0 +1,41 @@
...
+    },
+
+    data() {
+        return {
+            msgInherited: t`Inherited`,
+        };
+    },
+
+    methods: {
```
le0m:
```suggestion
    computed: {
```
These are good candidates for computed properties, instead of being methods. The difference is that the result of the function is cached based on the property used (ex. result of `isLocked` is updated only when `this.userRoles` or `this.role` change). If you change this, you can use them in template as normal properties (ex. `isLocked()` becomes `isLocked`)

resources/js/app/components/permissions/permissions.vue
```diff
@@ -0,0 +1,69 @@
...
+
+    name: 'permissions',
+
+    components: {
+        Permission:() => import(/* webpackChunkName: "permission" */'app/components/permission/permission'),
+    },
+
+    props: {
+        objectPerms: [],
+        userRoles: [],
```
le0m:
```suggestion
        objectPerms: {
            type: Array,
            default: () => ([]),
        },
        userRoles: {
            type: Array,
            default: () => ([]),
        },
```

le0m (approved):

PR #982, opened 2023-04-07
Show object title, module name and project name in `<title>`

paolosantarsiero:
This provides an update in pages `<title>`, introducing a new function `LayoutHelper::title`.
When `$object['attributes']['title']` is available, title starts with it.

Ref: https://chialab.atlassian.net/browse/ZEB-12

### DIFF
--- a/src/View/Helper/LayoutHelper.php
+++ b/src/View/Helper/LayoutHelper.php
@@ -107,7 +107,7 @@ class LayoutHelper extends Helper
         $title = '';
         if (!empty($currentModule) && !empty($currentModule['name'])) {
             $currentModuleName = $currentModule['name'];
-            if (!empty($object) && !empty($object['attributes']['title'])) {
+            if (Hash::check($object, 'attributes.title')) {
                 $title = $object['attributes']['title'] . ' | ';
             }
             $title = $title . $currentModuleName;
--- a/src/View/Helper/LayoutHelper.php
+++ b/src/View/Helper/LayoutHelper.php
@@ -96,6 +96,27 @@ class LayoutHelper extends Helper
     }
 
     /**
+     * Return title with object title and current module name
+     *
+     * @return string title with object title and current module name separated by '|'
+     */
+    public function title(): string
+    {
+        $currentModule = (array)$this->getView()->get('currentModule');
+        $object = (array)$this->getView()->get('object');
+        $title = '';
+        if (!empty($currentModule) && !empty($currentModule['name'])) {
+            $currentModuleName = $currentModule['name'];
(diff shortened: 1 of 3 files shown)


tests/TestCase/View/Helper/LayoutHelperTest.php
```diff
@@ -197,6 +197,75 @@ public function testModuleLink($expected, $name, array $viewVars = []): void
...
+    /**
+     * Test `title` method
+     *
+     * @param string $expected The expected title
+     * @param string $name The view name
+     * @param array $viewVars The view vars
+     * @dataProvider titleProvider()
+     * @covers ::title()
+     */
+    public function testTitle($expected, $name, array $viewVars = []): void
```
didoda:
```suggestion
    public function testTitle(string $expected, string $name, array $viewVars = []): void
```
You can specify types in signature ;)

src/View/Helper/LayoutHelper.php
```diff
@@ -95,6 +95,27 @@ public function messages(): bool
...
+     * @return string title with object title and current module name separated by '|'
+     */
+    public function title(): string
+    {
+        $currentModule = (array)$this->getView()->get('currentModule');
+        $object = (array)$this->getView()->get('object');
+        $title = '';
+        if (!empty($currentModule) && !empty($currentModule['name'])) {
+            $currentModuleName = $currentModule['name'];
+            if (!empty($object) && !empty($object['attributes']['title'])) {
```
didoda:
```suggestion
            if (Hash::check($object, 'attributes.title')) {
```
You can use `Cake\Utility\Hash::check` in here

didoda (approved):

src/View/Helper/LayoutHelper.php
```diff
@@ -95,6 +95,27 @@ public function messages(): bool
...
+        if (!empty($currentModule) && !empty($currentModule['name'])) {
+            $currentModuleName = $currentModule['name'];
+            if (Hash::check($object, 'attributes.title')) {
+                $title = $object['attributes']['title'] . ' | ';
+            }
+            $title = $title . $currentModuleName;
+        }
+
+        return $title;
+    }
```
didoda, line 111: This function could avoid the double `if`, using `Hash::get` like:
```
public function title(): string
{
    $module = (array)$this->getView()->get('currentModule');
    $name = (string)Hash::get($module, 'name');
    $object = (array)$this->getView()->get('object');
    $title = (string)Hash::get($object, 'attributes.title');
    
    return empty($title) ? $name : sprintf('%s | %s', $title, $name);
}
```

didoda (approved):

PR #983, opened 2023-04-12
Show resource change in edit

paolosantarsiero:
Show floppy icon after resource editing


resources/js/app/app.js
```diff
@@ -149,6 +149,7 @@ const _vueInstance = new Vue({
         // listen events emitted on this vue instance
         this.$on('filter-update-page-size', this.onUpdatePageSize);
         this.$on('filter-update-current-page', this.onUpdateCurrentPage);
+        this.$on('resource-changed', () => document.getElementById('module-icon').classList.add('resource-changed'));
```
didoda: I'd suggest to use a new function (`onResourceChanged`?) where you can check also the existence of element with id `module-icon`; if it doesn't exist, this line would provoke a javascript error

didoda (approved):

PR #988, opened 2023-04-20
Save folder permissions

didoda:
This provides the ability to save permissions on folders.

Requirement: https://github.com/bedita/bedita/pull/2005.

### DIFF
--- a/src/Controller/Admin/RolesController.php
+++ b/src/Controller/Admin/RolesController.php
@@ -56,8 +56,7 @@ class RolesController extends AdministrationBaseController
     public function save(): ?Response
     {
         $response = parent::save();
-        $roles = Hash::combine((array)$this->apiClient->get('/roles'), 'data.{n}.id', 'data.{n}.attributes.name');
-        Cache::write(self::CACHE_KEY_ROLES, $roles);
+        Cache::delete(self::CACHE_KEY_ROLES);
 
         return $response;
     }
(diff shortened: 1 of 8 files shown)


src/Controller/Admin/RolesController.php
```diff
@@ -41,4 +47,18 @@ class RolesController extends AdministrationBaseController
...
+    /**
+     * Save data
+     *
+     * @return \Cake\Http\Response|null
+     */
+    public function save(): ?Response
+    {
+        $response = parent::save();
+        $roles = Hash::combine((array)$this->apiClient->get('/roles'), 'data.{n}.id', 'data.{n}.attributes.name');
+        Cache::write(self::CACHE_KEY_ROLES, $roles);
```
fquffio: Since this cache is already being created elsewhere when needed, I'd just purge it in this action, so that we don't duplicate the logic: in case we ever need to change the cached value format, we have to change it both in `PermissionsTrait` and here.

To avoid this scenario, I'd just clear it here, and it will be written by PermsTrait the first time it's needed again.

```suggestion
        Cache::delete(self::CACHE_KEY_ROLES);
```

didoda: Cache invalidation should be performed in roles removal as well. So I added the `remove` method with a `Cache::delete(self::CACHE_KEY_ROLES)`.

fquffio (approved):

PR #990, opened 2023-04-26
Field info if admin

didoda:
This adds extra info in module view form fields, when user has role `admin`.

The info consists of json used to build the field, shown as tooltip when mouse is over the `i` box close to the field.

An example: 
image


didoda:
Not that useful.

PR #997, opened 2023-05-15
Object locked by parent(s)

didoda:
This provides an update in object view and index. When object parents permissions roles are all different from user's roles, object is considered "locked by parent(s)", meaning that the object:

 - cannot be deleted (in object view and index, no "remove" button)
 - cannot be moved to another position (in object view, some tree nodes checkboxes are disabled)
 - its `uname` cannot be changed (in object view, uname input text field is readonly)
 - its `status` cannot be changed (in object view, status radio is readonly)
 - UI: in the object view, the module box link has a lock icon background

### DIFF
--- a/src/View/Helper/PermsHelper.php
+++ b/src/View/Helper/PermsHelper.php
@@ -208,6 +208,9 @@ class PermsHelper extends Helper
 
     /**
      * Return true if object is locked by parents.
+     * When user is admin, return false.
+     * When user is not admin, return false if at least one parent is not locked for user.
+     * Return true if all parents are locked for user.
      *
      * @param string $id The object id
      * @return bool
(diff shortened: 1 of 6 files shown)


src/View/Helper/PermsHelper.php
```diff
@@ -201,4 +205,35 @@ public function userRoles(): array
...
+        }
+        $roles = $this->userRoles();
+        $locked = true;
+        foreach ($included as $data) {
+            if (count(array_intersect($roles, (array)Hash::get($data, 'meta.perms.roles'))) > 0) {
+                $locked = false;
+            }
+        }
+
+        return $locked;
```
le0m:
```suggestion
        foreach ($included as $data) {
            if (count(array_intersect($roles, (array)Hash::get($data, 'meta.perms.roles'))) > 0) {
                return false;
            }
        }

        return true;
```

PR #998, opened 2023-05-17
Attempt to improve locations autocomplete

edoardocavazza:
This PR introduces a change to the autocomplete algorithm for locations.

First of all, it uses pagination to collect at least 20 results. The current conditions are kept, and a new one is added to avoid duplicates among addresses (in case there are duplicates among the locations, which would still need to be cleaned).

### DIFF
--- a/resources/js/app/components/locations-view/location-view.js
+++ b/resources/js/app/components/locations-view/location-view.js
@@ -72,7 +72,7 @@ export default {
                         <autocomplete
                             autocomplete="none"
                             ref="address"
-                            class="autocomplete-address"
+                            class="autocomplete autocomplete-address"
                             :default-value="address"
                             :search="searchAddress"
                             base-class="autocomplete-address"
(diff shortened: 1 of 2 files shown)


nicolocarpignoli:
@edoardocavazza 

The fix is okay. However, the algorithm shows the results after a series of calls. You have to wait for all of them. Now, I don't know if this becomes a real problem in production, but in my local environment it kind of is: when the user starts typing, if he does not see any result in 5-6 seconds he might start typing again, and so on etc. etc. In fact never using existing locations.

To solve it easily, in my opinion it makes sense to insert a loader (already used elsewhere) to let the user know that he has to wait a moment to see if there are any results.

Translated with www.DeepL.com/Translator (free version)

nicolocarpignoli:
@edoardocavazza

The fix is okay. However, the algorithm shows the results after a series of calls. You have to wait for all of them. Now, I don't know if this becomes a real problem in production, but in my local environment it kind of is: when the user starts typing, if he does not see any result in 5-6 seconds he might start typing again, and so on etc. etc. In fact never using existing locations.

To solve it easily, in my opinion it makes sense to insert a loader (already used elsewhere) to let the user know that he has to wait a moment to see if there are any results.

edoardocavazza:
* Increased page size of API calls to 100 (less requests)
* Added spinner

PR #999, opened 2023-05-19
Use `async_jobs` endpoint

didoda:
This PR provides a change to use the new `/async_jobs` endpoint in data import page.

Import page link is shown to user only if `async_jobs` is allowed for `GET` or `POST` for the user (the system check `GET /home` response for that), and if there's a `Filters.import` configuration.

In import page:

 - form upload and submit section is shown if user is allowed for `POST /async_jobs` 
 - job list section is shown if user is allowed for `GET /async_jobs`

In the actual page, `/admin/async_jobs` endpoint is used: admin only can use it.
With this change, non-admin user will be able to import data as well, if API provides an endpoint_permission for manager and user role to do that.

API dependency: https://github.com/bedita/bedita/pull/2012

### DIFF
--- a/src/Utility/SchemaTrait.php
+++ b/src/Utility/SchemaTrait.php
@@ -18,11 +18,14 @@ use Authentication\Identity;
 use BEdita\SDK\BEditaClientException;
 use BEdita\WebTools\ApiClientProvider;
 use Cake\Cache\Cache;
+use Cake\Log\LogTrait;
 use Cake\Utility\Hash;
 use Psr\Log\LogLevel;
 
 trait SchemaTrait
 {
+    use LogTrait;
+
     /**
      * Getter for home endpoint metadata from user identity.
      *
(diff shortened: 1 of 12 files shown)


src/Utility/SchemaTrait.php
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * Getter for home endpoint metadata.
+     *
+     * @return array
+     */
+    public function getMeta(): array
+    {
+        try {
+            /** @var \Authentication\Identity|null $user */
+            $user = $this->Authentication->getIdentity();
```
batopa: Since it requires `AuthenticationComponent` maybe it's better to maintain this method in `ModulesComponent`. Loading component in controller action would skip `ModulesComponent::startup()` method if it's a problem.

If you want to use it as trait I would suggest to change the method signature as `getMeta(Identity $user)`, add `use LogTrait` and refactor `ModulesComponent::getMeta()` to call `getMeta($user)` of trait.

batopa (approved):

PR #1050, opened 2023-10-09
RolesModules and Admin index load data

didoda:
This provides an enhancement in Admin sections:

 - Admin all modules indexes: all data is retrieved (iterating over pagination result)
 - Admin Roles Modules: button save (and form) inside each block of permissions, to avoid an error of "too many variables in form"
 - User view - roles: load roles with limit 100 (it was 20), show a scrollbar when there are several roles

### DIFF
--- a/resources/js/app/components/relation-view/roles-list-view.vue
+++ b/resources/js/app/components/relation-view/roles-list-view.vue
@@ -52,7 +52,7 @@ export default {
     },
 
     async mounted() {
-        await this.loadObjects();
+        await this.loadObjects(true, {page_size:100});
         this.$nextTick(() => {
             let groups = this.groups;
             if (Object.keys(groups).length === 0) {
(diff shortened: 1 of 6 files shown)


resources/js/app/components/relation-view/roles-list-view.vue
```diff
@@ -119,3 +119,9 @@ export default {
     },
 }
 </script>
+<style>
+.roles-list-view {
```
nicolocarpignoli: minor: indentation

resources/js/app/components/relation-view/roles-list-view.vue
```diff
@@ -52,7 +52,7 @@ export default {
     },
 
     async mounted() {
-        await this.loadObjects();
+        await this.loadObjects(true, {page_size:100});
```
nicolocarpignoli: minor: spacing 

`{ page_size: 100 })`

PR #1052, opened 2023-10-12
POC DeeplTranslator

didoda:
This introduces `DeeplTranslator` class, that uses `babymarkt/deepl-php-lib`.

Setup example. In `config/app_local.php`:
```php
'Translator' => [
    'class' => '\App\Core\I18n\DeepLTranslator',
    'options' => [
        'auth_key' => '******************************',
    ],
],
```

When `Translator` is set, Manager UI provides "auto-translate" buttons in translations views.

Note: this is a POC. This logic will be moved to a separate plugin, with other translators implementations (i.e. AWSTranslator).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,6 +22,7 @@
     ],
     "require": {
         "php": ">=7.4",
+        "babymarkt/deepl-php-lib": "^1.0",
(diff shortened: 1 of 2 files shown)


didoda:
Only a POC

PR #1077, opened 2024-01-18
Use concrete object types in side Panel

didoda:
This provides a refactor to use concrete object types in object view side panel, when you open it (on "add objects" button click, for a relation, for example).

This also updates placeholders plugin to provide proper "available types" when you try to add placeholder and side panel opens.


src/Utility/Schema.php
```diff
@@ -21,18 +21,18 @@
...
 {
     /**
-     * Return unique alphabetically ordered right types from schema $schema
+     * Return unique alphabetically ordered object types from map
      *
-     * @param array $schema Schema data.
+     * @param array $relationsSchema The relations schema
      * @return array
      */
-    public static function rightTypes(array $schema): array
```
batopa: Since it is a `public` method I think it would be better to maintain this method and mark it as `@deprecated Will be dropped in 5.x version`.

src/View/Helper/SchemaHelper.php
```diff
@@ -299,16 +299,4 @@ public function sortable(string $field): bool
 
         return !in_array($type, ['array', 'object']);
     }
-
-    /**
-     * Return unique right types from schema "relationsSchema".
-     *
-     * @return array
-     */
-    public function rightTypes(): array
```
batopa, line 308: As the other comment, I would mark this method as deprecated without removing it

batopa (changes requested):

src/Utility/Schema.php
didoda: method restored as deprecated

src/View/Helper/SchemaHelper.php
didoda: method restored as deprecated

batopa (approved):

paolosantarsiero (approved):

PR #1111, opened 2024-05-06
Search items by id or uname in side panel

andresbautista:
This provides a change in side panel when adding related objects: you can search objects by ID or uname.

image

### DIFF
--- a/templates/Element/FilterBox/filter_box.twig
+++ b/templates/Element/FilterBox/filter_box.twig
@@ -1,7 +1,7 @@
 <div class="filter-box">
     {% if not hideFilter %}
     <div>
-        {{ element('FilterBox/filter_box_common') }}
+        {{ element('FilterBox/filter_box_common', { showFilterSearchByType }) }}
     </div>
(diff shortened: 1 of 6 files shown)


templates/Element/FilterBox/filter_box.twig
```diff
@@ -1,7 +1,7 @@
 <div class="filter-box">
     {% if not hideFilter %}
     <div>
-        {{ element('FilterBox/filter_box_common') }}
+        {{ element('FilterBox/filter_box_common', { showFilterSearchByType }) }}
```
didoda, line 4:
```suggestion
        {{ element('FilterBox/filter_box_common', { 'showFilterSearchByType': true }) }}
```

andresbautista: The variable `showFilterSearchByType` cannot take the value `true` inside `filter_box` because it inherits the value from the parent element, which in this case is `relations_add`.

didoda (approved):

PR #1124, opened 2024-05-28
Object history: view all changes

didoda:
This fixes a limit in object history view, to max 100 elements, by providing multiple api calls to obtain all history items per object.

### DIFF
--- a/resources/js/app/components/history/history-changes.vue
+++ b/resources/js/app/components/history/history-changes.vue
@@ -88,9 +88,7 @@ export default {
         let count = 0;
         while (!done) {
             const historyRes = await fetch(`${baseUrl}${this.object.type}/history/${this.object.id}/${page}`, options);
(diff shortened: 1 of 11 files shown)


resources/js/app/components/history/history-changes.vue
```diff
@@ -1,49 +1,119 @@
...
+        };
+
+        this.isLoading = true;
+        this.canSave = this.cansave;
+        let done = false;
+        let page = 1;
+        let count = 0;
+        while (!done) {
+            const historyRes = await fetch(`${baseUrl}${this.object.type}/history/${this.object.id}/${page}`, options);
+            console.log(historyRes);
```
nicolocarpignoli: Here and two lines below, maybe we can remove the logs

nicolocarpignoli (approved):

PR #1130, opened 2024-06-03
Date ranges data: fix params on save

didoda:
We noticed that sometimes values in `params` of a date range are not consistent with the date range interval and internal data.
This provides a way to fix data using a new class `DateRangesTools` (method `prepare`).

All imagined cases are handled in unit tests.

Note: params can be `null` or a json object.
Accepted keys are: `all_day`, `every_day` and `weekdays`.

 - all_day: if checked, then `true`
 - every_day: if checked, then `true`
 - weekdays: string array of checked days

Params saved examples follow.

All day and every day:
```json
{
    "all_day": true,
    "every_day": true
}
```

All day and some days only
```json
{
    "all_day": true,
    "weekdays": ["sunday", "monday"]
}
```

Bonus: Undo button on remove.

### DIFF
--- a/src/Utility/DateRangesTools.php
+++ b/src/Utility/DateRangesTools.php
@@ -70,7 +70,7 @@ class DateRangesTools
 
                     continue;
                 }
-                // all_day is true, no need for weekdays
+                // all_day is true, no need for weekdays, and every_day MUST be true
                 $item['params'] = json_encode([
                     'all_day' => true,
                     'every_day' => true,
--- a/src/Utility/DateRangesTools.php
+++ b/src/Utility/DateRangesTools.php
@@ -105,7 +105,7 @@ class DateRangesTools
         if (!empty($weekdays)) {
             $params['weekdays'] = $weekdays;
         }
-        if (($everyDay || count($weekdays) === 7) || (!$everyDay && count($weekdays) === 0)) {
+        if (($everyDay && count($weekdays) === 7) || (!$everyDay && count($weekdays) === 0)) {
             $params['every_day'] = 'on';
             $params = array_filter(
                 $params,
--- a/src/Utility/DateRangesTools.php
+++ b/src/Utility/DateRangesTools.php
@@ -56,7 +56,7 @@ class DateRangesTools
             if (empty(Hash::get($item, 'params'))) {
                 continue;
             }
-            $item['params'] = self::parseParams((array)Hash::get($item, 'params'), self::isOneDayRange($item));
+            $item['params'] = self::parseParams((array)Hash::get($item, 'params'), self::isOneDayRange((array)$item));
         }
         $dateRanges = array_values($dateRanges);
 
--- a/src/Utility/DateRangesTools.php
+++ b/src/Utility/DateRangesTools.php
@@ -142,22 +142,4 @@ class DateRangesTools
 
         return empty($data) ? null : $data;
     }
-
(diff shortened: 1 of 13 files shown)


src/Utility/DateRangesTools.php
```diff
@@ -0,0 +1,84 @@
...
+            $allDay = Hash::get($params, 'all_day');
+            if (!empty($sd) && empty($ed)) {
+                if (!array_key_exists('all_day', $params) || $allDay === false) {
+                    $item['params'] = null;
+
+                    continue;
+                }
+                // all_day is true, no need for weekdays
+                $item['params'] = json_encode([
+                    'all_day' => true,
```
nicolocarpignoli: Looking at the data on the Manager, are we sure "all_day" when active is set on "true"? It looks like it is set to "on" string. Although that might be a later modify. If not, we should set it to "on" here.

src/Utility/DateRangesTools.php
```diff
@@ -0,0 +1,84 @@
...
+            $params = (string)Hash::get($item, 'params');
+            if (empty($params)) {
+                continue;
+            }
+            $params = json_decode($params, true);
+            $sd = (string)Hash::get($item, 'start_date');
+            $ed = (string)Hash::get($item, 'end_date');
+            $allDay = Hash::get($params, 'all_day');
+            if (!empty($sd) && empty($ed)) {
+                if (!array_key_exists('all_day', $params) || $allDay === false) {
```
nicolocarpignoli: The key we should check here is "every_day": "all_day" is viable even if we had only the start date.

src/Utility/DateRangesTools.php
```diff
@@ -0,0 +1,84 @@
...
+                    $item['params'] = null;
+
+                    continue;
+                }
+                // all_day is true, no need for weekdays
```
nicolocarpignoli: this should be "every_day"

didoda: I applied a refactor to consider all cases / code has changed...

src/Utility/DateRangesTools.php
```diff
@@ -0,0 +1,139 @@
...
+    {
+        // remove all_day and every_day if not needed
+        $params = self::filterNotOn($params, 'all_day');
+        $params = self::filterNotOn($params, 'every_day');
+        // one day range
+        $allDayOn = Hash::get($params, 'all_day') === 'on';
+        if ($oneDayRange) {
+            return $allDayOn ? ['all_day' => 'on', 'every_day' => 'on'] : null;
+        }
+        // if multi day, all_day has no sense: remove it
```
nicolocarpignoli: I do not agree with this statement and this logic, I think we should let the user set "all_day" for single dates but also for intervals, where the user may not wish to specify a precise time for the start and end date.

didoda: ok... but if you look how it works on BEM, if you check "all day", the UI forces the two input fields (days) to be the same day... According to this logic, "all_day" makes sense only for "one day" ranges

didoda: I'm wrong... sorry for the misunderstanding... I'll handle properly "all_day"

nicolocarpignoli: I can't reproduce the case you are referring to. The UI (before this PR) admit that:
image

PR #1141, opened 2024-06-17
Check image resolution: warning when image could not load

didoda:
The image resolution check, on image upload, sometimes cannot verify the image resolution, because the image "onload" fails.
In these cases, a proper warning pops up.

### DIFF
--- a/resources/js/app/helpers/view.js
+++ b/resources/js/app/helpers/view.js
@@ -106,6 +106,12 @@ export default {
                 const maxY = parts[1];
                 const img = new Image();
                 img.src = window.URL.createObjectURL(file);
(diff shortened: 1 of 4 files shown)


resources/js/app/helpers/view.js
```diff
@@ -106,6 +106,12 @@ export default {
                 const maxY = parts[1];
                 const img = new Image();
                 img.src = window.URL.createObjectURL(file);
+                img.onerror = (e) => {
+                    window.URL.revokeObjectURL(img.src);console.error('error', e);
```
nicolocarpignoli: I would add a newline before the console.error. Just for consistency

PR #1174, opened 2024-08-13
Clone object translations

### LINKED ISSUE #957: Clone include translations
When cloning an object, there should be an option to include translations.

Tip: better providing a "clone translations" directly in api call in https://github.com/bedita/bedita: see https://github.com/bedita/bedita/issues/2017

didoda:
This resolves https://github.com/bedita/manager/issues/957 by introducing the possibility to clone translations when cloning an object.

image

### DIFF
--- a/resources/js/app/components/dialog/dialog.js
+++ b/resources/js/app/components/dialog/dialog.js
@@ -79,8 +79,8 @@ export const Dialog = Vue.extend({
     },
     methods: {
         callbackOptions(title, unique) {
-            const relations = this.checks?.filter(c => c?.name === 'relations')?.[0]?.value || false;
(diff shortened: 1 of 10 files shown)


resources/js/app/components/dialog/dialog.js
```diff
@@ -70,13 +72,21 @@ export const Dialog = Vue.extend({
...
-            checkLabel: '',
+            checks: [],
             loading: false,
             unique: [],
         };
     },
     methods: {
+        callbackOptions(title, unique) {
+            const relations = this.checks?.filter(c => c?.name === 'relations')?.[0]?.value || false;
+            const translations = this.checks?.filter(c => c?.name === 'translations')?.[0]?.value || false;
```
paolosantarsiero: You can use `.find` instead `.filter` so you don't have to get first element of array

PR #1205, opened 2024-11-26
Clone relations with meta (and relation parameters)

andresbautista:
Introduces the possibility of passing the meta of a relationship when it is cloned


src/Controller/Component/CloneComponent.php
```diff
@@ -222,6 +222,7 @@ public function relation(string $sourceId, string $type, string $relation, strin
                 [
                     'id' => (string)Hash::get($obj, 'id'),
                     'type' => (string)Hash::get($obj, 'type'),
+                    'meta' => (array)Hash::get($obj, 'meta'), // <== questa riga
```
didoda:
```suggestion
                    'meta' => (array)Hash::get($obj, 'meta'),
```

didoda (approved):

PR #1214, opened 2025-01-03
Avoid sorting custom props from model

stefanorosanelli:
This PR introduces a fix on how `sortable` properties in modules index are handled. 
Instead of relaying on a configuration we check if the property type is sortable (property must be scalar, not a vector or an object) and if it's not a custom property since those properties are not sortable in BEdita API right now.

### DIFF
--- a/src/Controller/Component/SchemaComponent.php
+++ b/src/Controller/Component/SchemaComponent.php
@@ -365,6 +365,41 @@ class SchemaComponent extends Component
     }
 
     /**
+     * Retrieve list of custom properties for a given object type.
+     *
+     * @param string $type Object type name.
+     * @return array
+     */
+    public function customProps(string $type): array
+    {
+        return (array)Cache::remember(
+            CacheTools::cacheKey(sprintf('custom_props_%s', $type)),
+            function () use ($type) {
+                return $this->fetchCustomProps($type);
+            },
+            self::CACHE_CONFIG
+        );
+    }
+
+    /**
+     * Fetch custom properties for a given object type.
+     *
+     * @param string $type Object type name.
+     * @return array
+     */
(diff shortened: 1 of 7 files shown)


src/Controller/Component/SchemaComponent.php
```diff
@@ -364,6 +364,44 @@ public function descendants(string $type): array
...
+     */
+    protected function fetchCustomProps(string $type): array
+    {
+        if ($this->getConfig('internalSchema')) {
+            return []; // internal resources don't have custom properties
+        }
+        $query = [
+            'fields' => 'name',
+            'filter' => ['object_type' => $type, 'type' => 'dynamic'],
+            'page_size' => 100,
```
didoda: 100 custom properties is probably enough. However, to be completely safe, this part should iterate over pagination results.

This function could be:
```php
protected function fetchCustomProps(string $type): array
{
    if ($this->getConfig('internalSchema')) {
        return []; // internal resources don't have custom properties
    }
    $customProperties = [];
    $pageCount = $page = 1;
    while ($page <= $pageCount) {
        $query = [
            'fields' => 'name',
            'filter' => ['object_type' => $type, 'type' => 'dynamic'],
            'page' => $page,
            'page_size' => 100,
        ];
        $response = ApiClientProvider::getApiClient()->get('/model/properties', $query);
        $customProperties = array_merge($customProperties, (array)Hash::extract($response, 'data.{n}.attributes.name'));
        $pageCount = (int)Hash::get($response, 'meta.pagination.page_count');
        $page++;
    }
    sort($customProperties);

    return $customProperties;
}
```

stefanorosanelli: Done

didoda (approved):

PR #1257, opened 2025-03-31
Encode placeholder params using utf-8

paolosantarsiero:
This PR fix enconding of placeholders params using input as a utf-8 string, because method `atob` encode only Latin-1 (ISO-8859).


resources/js/app/components/placeholder-list/placeholder-params.vue
```diff
@@ -52,6 +52,7 @@
 </template>
 <script>
 import { EventBus } from 'app/components/event-bus';
+import { placeholderParamsDecoded, placeholderParamsEncode } from 'app/helpers/placeholder-encoding';
```
didoda: `atob` stands for `ascii to binary conversion` and `btoa` stands for `binary to ascii conversion`.
We could rename these new functions to something like `asciiToBinaryUtf8` (or `atobUtf8`) and `binaryToAsciiUtf8` (or `btoaUtf8`).
Or the functions could be `encode***` and `decode***`.

paolosantarsiero: If we decide to change these names we can move all methods to text helpers. I will to do

paolosantarsiero: If we decide to change these names we can move all methods to text helpers. I will to do

didoda (approved):

PR #1432, opened 2026-07-28
[6.x] fix: Remove conflict in composer.json

didoda:
This removes a "conflict" condition in composer.json, to allow using recent version of slevomat coding standard.


didoda:
wrong PR

